PR #5, opened 2015-01-28
Iaguis/library

### LINKED ISSUE #3: docker2aci: create a library that can be imported
As well as providing a command line tool, `docker2aci` should be usable as a library that other projects can leverage to work with docker images.

For example, rocket might import `docker2aci` and use it to be able to fetch/run directly from Docker registry URLs.

(The subtext here is that there should be nothing specific to rocket or any other implementation in this project itself)

iaguis:
This pull request makes docker2aci a library so it can be imported and used.
Fixes #3


lib/docker2aci.go
```diff
@@ -127,146 +178,191 @@ func GetRepoData(indexURL string, remote string) (*RepoData, error) {
...
-	return ancestry, nil
+	return jsonBytes, imageSize, nil
 }
 
-func setAuthToken(req *http.Request, token []string) {
-	if req.Header.Get("Authorization") == "" {
-		req.Header.Set("Authorization", "Token "+strings.Join(token, ","))
+func getRemoteLayer(imgID, registry string, repoData *RepoData, imgSize int64) (io.ReadCloser, error) {
+	client := &http.Client{}
+	req, err := http.NewRequest("GET", registry+"images/"+imgID+"/layer", nil)
```
philips: path.Join

lib/docker2aci.go
```diff
@@ -336,83 +433,27 @@ func GenerateManifest(layerData DockerImageData, dockerURL *DockerURL) (*schema.
...
-	imageName := appString
-	appParts := strings.Split(appString, ":")
-
-	if len(appParts) > 1 {
-		tag = appParts[len(appParts)-1]
-		imageNameParts := appParts[0 : len(appParts)-1]
-		imageName = strings.Join(imageNameParts, ":")
+func parseDockerUser(dockerUser string) (string, string) {
+	if dockerUser == "" {
+		return "0", "0"
```
philips: this deserves a comment :-/

lib/util.go
```diff
@@ -0,0 +1,68 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("Error reading layer (gzip): %v", err)
+		}
+		return gz, nil
+	case Uncompressed:
+		return bufR, nil
+	default:
+		return nil, fmt.Errorf("Unknown layer format")
+	}
+}
```
philips: Hrm, these things are already in the aci libraries: http://godoc.org/github.com/appc/spec/aci#NewCompressedReader

iaguis: Oh, didn't see those. I'll use them where it makes sense and remove this.

main.go
```diff
@@ -0,0 +1,88 @@
...
+
+	"github.com/appc/docker2aci/lib"
+	"github.com/coreos/rocket/cas"
+)
+
+const (
+	rocketDir = "/var/lib/rkt"
+)
+
+var flagImport = flag.Bool("import", false, "Import ACI images to the rocket store")
```
strangemonad: while certainly convenient, I feel it's a little odd that this introduces a dependency on rocket (a specific implementation of the appc spec). I feel it's also more unix-y if the rocker import and docker2aci remain orthogonal. I personally wouldn't view it as much of a burden to chain this command with a command to import the generated ACI

(again, I'm not very familiar with the appc / rocket projects yet, just providing CLI design feedback as an excited user)

jonboulle: You beat me to it :-). This absolutely should not be here, there will be no Rocket related code in appc. It was just from early prototyping.

What's here now will end up in Rocket itself - tracking ticket: https://github.com/coreos/rocket/issues/145

main.go
```diff
@@ -0,0 +1,88 @@
...
+
+	return rocketImageID, nil
+}
+
+func main() {
+	flag.Parse()
+	args := flag.Args()
+
+	if len(args) != 1 {
+		fmt.Println("Usage: docker2aci [--import] [REGISTRYURL/]IMAGE_NAME[:TAG]")
```
strangemonad: flag and usage will give you a lot of this already and make sure that your command feels like a consistent golang cmd:

a good example of a cmd runner is here:
http://golang.org/src/cmd/go/main.go

a shorter example of cmd flags is here:
https://gobyexample.com/command-line-flags

main.go
strangemonad: @jonboulle I'm assuming coreos/rocket#145 will just re-use this library? I certainly feel a standalone docker2aci cmd is useful as part of the appc project (separate from rocket)

jonboulle: @strangemonad yep, right on both counts! Ultimately I could anticipate other projects using this library as well.

philips:
Should we close this one since the follow-on PR has most everything?

PR #6, opened 2015-01-28
Iaguis/squash

### LINKED ISSUE #2: docker2aci: support squashing images
In the case of an image with multiple layers, it would be nice if `docker2aci --squash` worked; rather than generating ACIs for each layer and turning them into dependencies, it would simply merge the filesystems of all layers into a single ACI.

iaguis:
This PR implements squash functionality for ACI images. The last 3 commits are the interesting ones, the rest is covered by #5.

Fixes #2


iaguis:
Updated PR

lib/docker2aci.go
```diff
@@ -48,41 +47,124 @@ type DockerURL struct {
...
+	}
 
-	for _, ep := range headers {
-		endpointsList := strings.Split(ep, ",")
-		for _, endpointEl := range endpointsList {
-			endpoints = append(
-				endpoints,
-				// TODO(iaguis) discover if httpsOrHTTP
-				fmt.Sprintf("https://%s/v1/", strings.TrimSpace(endpointEl)))
+	layersOutputDir := "."
```
jonboulle: this should just be outputDir no?

lib/docker2aci.go
```diff
@@ -48,41 +47,124 @@ type DockerURL struct {
...
+// If the squash flag is true, it squashes all the layers in one file and
+// places this file in outputDir; if it is false, it places every layer in its
+// own ACI in outputDir.
+// It returns the list of generated ACI paths.
+func Convert(dockerURL string, squash bool, outputDir string) ([]string, error) {
+	parsedURL := parseDockerURL(dockerURL)
+
+	repoData, err := getRepoData(parsedURL.IndexURL, parsedURL.ImageName)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error getting repository data: %v\n", err)
```
jonboulle: Generally speaking, don't print to stderr for a library function like this, just return the error and let the caller handle it. (You can still wrap the err if needed to make things clearer, e.g. `return fmt.Errorf("error getting repository data: %v", err)`)

lib/docker2aci.go
```diff
@@ -127,146 +209,203 @@ func GetRepoData(indexURL string, remote string) (*RepoData, error) {
...
 	if err != nil {
-		return "", fmt.Errorf("Failed to get Image ID: %s, URL: %s", err, req.URL)
+		return "", fmt.Errorf("Error creating dir: %v", err)
 	}
+	defer os.RemoveAll(tmpDir)
 
-	setAuthToken(req, repoData.Tokens)
-	setCookie(req, repoData.Cookie)
-	res, err := client.Do(req)
+	layerDest := tmpDir + "/layer"
```
jonboulle: path.Join

jonboulle: filepath.Join

lib/docker2aci.go
```diff
@@ -48,41 +47,124 @@ type DockerURL struct {
 	Tag       string
 }
 
-type Compression int
+type SquashAcc struct {
```
jonboulle: What is Acc?

lib/docker2aci.go
```diff
@@ -48,41 +47,124 @@ type DockerURL struct {
...
+		if err := squashLayers(aciLayerPaths, squashedImagePath); err != nil {
+			fmt.Fprintf(os.Stderr, "Error squashing image: %v\n", err)
+		}
+		aciLayerPaths = []string{squashedImagePath}
+	}
+
+	return aciLayerPaths, nil
+}
+
+func parseDockerURL(arg string) *DockerURL {
```
jonboulle: does this exist in upstream docker - should we just leverage it from there?

lib/docker2aci.go
```diff
@@ -127,146 +209,203 @@ func GetRepoData(indexURL string, remote string) (*RepoData, error) {
...
 }
 
-func GetRemoteImageJSON(imgID, registry string, repoData *RepoData) ([]byte, int, error) {
+func getImageIDFromTag(registry string, appName string, tag string, repoData *RepoData) (string, error) {
 	client := &http.Client{}
-	req, err := http.NewRequest("GET", registry+"images/"+imgID+"/json", nil)
+	req, err := http.NewRequest("GET", "https://"+path.Join(registry, "repositories", appName, "tags", tag), nil)
 	if err != nil {
-		return nil, -1, err
+		return "", fmt.Errorf("Failed to get Image ID: %s, URL: %s", err, req.URL)
```
jonboulle: errors should always start lowercase (in anticipation of being printed by callers with a prefix)

lib/docker2aci.go
```diff
@@ -127,146 +209,203 @@ func GetRepoData(indexURL string, remote string) (*RepoData, error) {
...
 	}
 	defer res.Body.Close()
 
 	if res.StatusCode != 200 {
-		return nil, -1, fmt.Errorf("HTTP code: %d, URL: %s", res.StatusCode, req.URL)
+		return "", fmt.Errorf("HTTP code: %d. URL: %s", res.StatusCode, req.URL)
 	}
 
-	imageSize := -1
+	jsonString, err := ioutil.ReadAll(res.Body)
```
jonboulle: For short-lived/narrowly scoped variables, more idiomatic to use short names; "j" or "b" would be fine here

lib/docker2aci.go
```diff
@@ -445,166 +520,154 @@ func WriteACI(layer io.Reader, manifest []byte, output string) error {
...
-		hdr.Name = "rootfs/"+hdr.Name
-		if err := trw.WriteHeader(hdr); err != nil {
-			return fmt.Errorf("Error writing header: %v", err)
+		if hdr.Name == "./" {
+			continue
 		}
-		if _, err := io.Copy(trw, tr); err != nil {
-			return fmt.Errorf("Error copying file into the tar out: %v", err)
+		// FIXME(iaguis) although unlikely, a file named like "/what.wh.ever should be legal
+		if strings.Index(hdr.Name, ".wh.") != -1 {
```
jonboulle: strings.Contains (although I would rather this TODO was fixed :P)

lib/docker2aci.go
```diff
@@ -48,41 +47,124 @@ type DockerURL struct {
 	Tag       string
 }
 
-type Compression int
+type SquashAcc struct {
+	OutWriter *tar.Writer
```
jonboulle: Probably simpler to just embed this as an anonymous struct.

lib/docker2aci.go
iaguis: Ooops, good catch. Thanks!

lib/docker2aci.go
iaguis: Forgot to change it, thanks

lib/docker2aci.go
iaguis: It stands for accumulator, because it accumulates all the layers' files, filenames and manifests. I can rename it to SquashAccumulator if that makes it clearer.

lib/docker2aci.go
iaguis: Yes, there're a couple of functions we can use

lib/docker2aci.go
iaguis: OK

lib/docker2aci.go
iaguis: OK

lib/docker2aci.go
iaguis: I'll fix it :)

lib/docker2aci.go
iaguis: I just learned that this existed :)

philips:
I think we should just move this and keep moving forward. /cc @jonboulle

sgotti:
@iaguis I was looking at the patchset. Wouldn't be better to reuse some existing code? Right now I splitted the acirenderer patch in two parts (coreos/rocket#464 and coreos/rocket#465) and the first one should be used also by docker2aci and doesn't need any previous PR.

I was thinking of this code path:

*) Create an ACI for every layer, specifying in its manifest the parent layer as a dependency.
*) Save these ACIs in a "store" implementation that satisfies to ACIProvider interface in (coreos/rocket#464) (like a cas or another easier implementation)
*) Compose the dependency list.
*) Use acirenderer (coreos/rocket#464) to render the squashed aci
*) Create the tar.

iaguis:
I'd like to reuse your code. Right now I see two problems though:

* acirenderer depends on rocket. It's only ptar so it should be easy to decouple
* acirenderer extracts the ACIs to disk. This simplifies everything but it requires root which I was trying to avoid.

Any thoughts on that? Maybe requiring root is not a big deal?

jonboulle:
> acirenderer depends on rocket. It's only ptar so it should be easy to decouple

Yep, no need for this to live in Rocket itself

> acirenderer extracts the ACIs to disk. This simplifies everything but it requires root which I was trying to avoid.

Hmm, why does it need root (a priori of being applied to a filesystem that needs root)?

jonboulle:
@iaguis @sgotti I would love to see the acirenderer code reused if at all possible, but I also think we should probably just land this now and follow up with that.

Do you think it makes sense for that code to move "upstream" to appc/spec or so?

iaguis:
> Hmm, why does it need root (a priori of being applied to a filesystem that needs root)?

Extracting files with the correct permissions/owners or with node files requires root since a regular user can't make the appropriate calls (chown, chmod...)

> I also think we should probably just land this now and follow up with that.

I agree

jonboulle:
Hah, OK, I was thinking completely backwards

sgotti:
@jonboulle @iaguis  You're right!

In the meantime I just changed acirenderer with a better algo so now it will create a list of files to extract for every aci.
This list can be used by another function to directly create a tar (like needed by docker2aci, or to extract it to a filesystem, like the original acirenderer). 

The extraction part is in its file (extract.go) so this can also be splitted and the acirenderer core (acirenderer.go and the tests) can be moved.
@jonboulle appc/spec can be ok as the place to move it.

PTAL (coreos/rocket#464)

iaguis:
That's awesome! I'll try to use it in docker2aci and let you know any issues I find.

philips:
I just found this library: https://github.com/winchman/libsquash

jonboulle:
Landing this and we can roll forward with the follow ups: #9 #10

PR #13, opened 2015-02-07
Remove docker dependencies

iaguis:
We depended on docker packages to use docker's json manifest structures
and to parse the registry URLs but they're in packages `parsers`,
`registry` and `runconfig` which are quite general and have a lot of
dependencies. This caused Godep to download a big chunk of docker's code
which is not desirable when integrating docker2aci with rocket, for
example.

This commit extracts the needed docker functions and structures so we
don't import docker packages anymore.

### DIFF
--- a/lib/docker2aci.go
+++ b/lib/docker2aci.go
@@ -20,39 +20,9 @@ import (
 	"github.com/appc/spec/aci"
 	"github.com/appc/spec/schema"
 	"github.com/appc/spec/schema/types"
-	"github.com/docker/docker/pkg/parsers"
-	"github.com/docker/docker/registry"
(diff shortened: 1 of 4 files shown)


jonboulle:
I would like to see the types that are upstreamed from Docker explicitly called out as such (and distinguished from the local docker2aci types). Aside from that, LGTM!

iaguis:
Only the upstreamed types start by Docker now, is this enough or should I be more explicit?

jonboulle:
@iaguis can you just add a brief comment describing it too?

iaguis:
Done

jonboulle:
:+1:

PR #20, opened 2015-02-11
Handle Entrypoint and Cmd correctly

### LINKED ISSUE #14: set default exec from ENTRYPOINT and CMD
Dockerfiles have two parameters controlling what gets executed: [`ENTRYPOINT`](https://docs.docker.com/reference/builder/#entrypoint) and [`CMD`](https://docs.docker.com/reference/builder/#cmd). There is a somewhat complicated relationship between the two as they both support _exec_ and _shell_ forms and `CMD` has a third form which just results in being passed as arguments to `ENTRYPOINT`.
 
We should just follow the same logic that Docker does for resolving these into the actual default command line to be executed, and then set that as the app.exec.cmd in our created image manifest.

iaguis:
Use Entrypoint + Cmd when both are present or just Cmd when Entrypoint
is not present.

Since non-absolute paths are not allowed by the spec, if the resulting
command is not absolute we fall back to running it with "/bin/sh -c".

Fixes #14

### DIFF
--- a/lib/docker2aci.go
+++ b/lib/docker2aci.go
@@ -417,12 +417,7 @@ func generateManifest(layerData DockerImageData, dockerURL *ParsedDockerURL) (*s
 	genManifest.Labels = labels
 
 	if dockerConfig != nil {
-		var exec types.Exec
-		if len(dockerConfig.Cmd) > 0 {
-			exec = types.Exec(dockerConfig.Cmd)
-		} else if len(dockerConfig.Entrypoint) > 0 {
-			exec = types.Exec(dockerConfig.Entrypoint)
-		}
+		exec := getExecCommand(dockerConfig.Entrypoint, dockerConfig.Cmd)
 		if exec != nil {
 			user, group := parseDockerUser(dockerConfig.User)
 			app := &types.App{Exec: exec, User: user, Group: group}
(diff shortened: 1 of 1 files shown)


lib/docker2aci.go
```diff
@@ -447,6 +442,20 @@ func generateManifest(layerData DockerImageData, dockerURL *ParsedDockerURL) (*s
 	return genManifest, nil
 }
 
+func getExecCommand(entrypoint []string, cmd []string) types.Exec {
+	var command []string
+	if entrypoint == nil && cmd == nil {
+		return nil
+	}
+	command = append(entrypoint, cmd...)
+	// non-absolute paths are not allowed, fallback to "/bin/sh -c command"
```
jonboulle: Starting to feel like we should have a means to provide warnings during the conversion process when things like this happen, thoughts?  (In the real world this would _probably_ very rarely cause problems since having /bin/sh in Docker images is a really common pattern, but would be good practise to give people _some_ hint that we are doing this kind of translation and the produced image is potentially non-functional)

iaguis: We could warn with something like `Exec is not an absolute path, using "/bin/sh -c EXEC". The resulting ACI could be non-functional`.

This will be printed for each layer that we convert so it might be too verbose, especially if we're squashing. We could accumulate the warnings and print only one of each at the end. Something like:

```
Warnings found when converting the image. The resulting ACI could be non-fuctional:
        Exec is not an absolute path, using "/bin/sh -c EXEC"
        Docker User is empty, assuming root
```

jonboulle: Yeah, printing is less than ideal since this is a library. Maybe we could allow callers to pass in something to record warnings. Not sure.

jonboulle: Let's follow up with that later, filed #24

lib/docker2aci.go
```diff
@@ -734,8 +743,9 @@ func mergeManifests(manifests []schema.ImageManifest) schema.ImageManifest {
 		manifest.Labels = append(manifest.Labels[:layerIndex], manifest.Labels[layerIndex+1:]...)
 	}
 
+	n := strings.LastIndex(manifest.Name.String(), "-")
 	// this can't fail because the old name is legal
-	nameWithoutLayerID, _ := types.NewACName(strings.Split(manifest.Name.String(), "-")[0])
```
jonboulle: huh, why was this necessary?

iaguis: Every layer's name has the docker layerID at the end, this strips it from the squashed image name. It was wrong because the image name can have dashes.

iaguis: I'll put it in another PR

iaguis:
Since we have #24 now, I'm merging this.

PR #21, opened 2015-02-11
docker2aci: remove import cmd

### LINKED ISSUE #11: remove rocket-specific import/logic
docker2aci should not have any dependency on code in coreos/rocket; instead we should see the inverse, where tools like Rocket leverage docker2aci to fetch/run Docker images (e.g. https://github.com/coreos/rocket/issues/145). So this ticket is to track the removal of the [current dependency](https://github.com/appc/docker2aci/blob/bb72ee53c79f7c863a694a306ab47fe22fb80e5f/main.go#L10) within the docker2aci CLI on a Rocket package.

It's possible that rather than removing the _import_ feature of docker2aci entirely, it could instead move to a more generic model - for example, the ACI Registry proposed in https://github.com/coreos/rocket/pull/464. I will leave that to the implementer to explore; the main point here is that the Rocket dep goes away  :-)

iaguis:
It introduces a dependency on rocket and it should be handled by another
tool.

Fixes #11

### DIFF
--- a/main.go
+++ b/main.go
@@ -1,25 +1,20 @@
 package main
 
 import (
-	"bufio"
 	"flag"
 	"fmt"
 	"os"
 
 	"github.com/appc/docker2aci/lib"
-	"github.com/coreos/rocket/cas"
 )
 
 const (
 	rocketDir = "/var/lib/rkt"
 )
 
-var (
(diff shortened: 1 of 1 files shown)


jonboulle:
Lgtm, thanks

PR #25, opened 2015-02-13
Fix squashed image name

iaguis:
We were taking the name without layerID as the part before the first
dash. This is wrong because a name can have dashes.

Fix it by taking the part before the last dash.


alban:
LGTM

iaguis:
Thanks!

PR #28, opened 2015-02-23
Iaguis/acirenderer

### LINKED ISSUE #10: use acirenderer?
https://github.com/coreos/rocket/pull/464 introduces a library for rendering ACIs from their embedded rootfs and dependencies. Following the discussion in https://github.com/appc/docker2aci/pull/6 , we should be able to split out this library and leverage it in docker2aci itself. This will probably be blocked on the library moving out of Rocket and upstream into the appc org somewhere (appc/spec? appc/tools?), since it is a general-purpose and not Rocket specific.

iaguis:
Once appc/spec#216 is merged, we can use the acirenderer lib to squash images.

Resolves #10

### DIFF
--- a/lib/conversion_store.go
+++ b/lib/conversion_store.go
@@ -1,0 +1,99 @@
+package docker2aci
+
+import (
+	"fmt"
+	"hash"
+	"io"
+	"io/ioutil"
+	"os"
+
+	"github.com/appc/spec/aci"
+	"github.com/appc/spec/schema"
+	"github.com/appc/spec/schema/types"
+)
+
+const (
+	hashPrefix = "sha512-"
+)
+
+type aciInfo struct {
+	path          string
+	key           string
+	ImageManifest *schema.ImageManifest
+}
+
+// ConversionStore is an simple implementation of the acirenderer.ACIRegistry
+// interface. It stores the Docker layers converted to ACI so we can take
+// advantage of acirenderer to generate a squashed ACI Image.
+type ConversionStore struct {
+	acis map[string]*aciInfo
+}
+
+func NewConversionStore() *ConversionStore {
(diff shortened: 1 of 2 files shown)


iaguis:
Now that acirenderer is merged, we can merge this. Please review.

lib/docker2aci.go
```diff
@@ -610,112 +621,96 @@ func writeManifest(outputWriter *tar.Writer, manifest schema.ImageManifest) erro
...
+					return fmt.Errorf("error writing header: %v", err)
+				}
+				if _, err := io.Copy(outputWriter, t.TarStream); err != nil {
+					return fmt.Errorf("error copying file into the tar out: %v", err)
+				}
+			}
+			return nil
+		}); err != nil {
+			return err
+		}
```
philips: not a huge fan of this error handling style, could you pull the walker out into a named function?

lib/conversion_store.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+	f, err := os.Open(aci.path)
+	if err != nil {
+		return nil, fmt.Errorf("error opening aci: %s", aci.path)
+	}
+
+	return f, nil
+}
+
+func (ms *ConversionStore) ResolveKey(key string) (string, error) {
```
philips: I want to rename all of these "key string" things to ID, I will do that against the spec here in a sec.

philips:
overall, lgtm besides the nits. If those are addressed feel free to merge.

lib/conversion_store.go
philips: thinking about it more the key naming is fine.

lib/docker2aci.go
iaguis: While I was at it I did the same for the other walker

PR #29, opened 2015-03-10
Fix handling of layers with no entrypoint or command

krobertson:
Fixes an issue with an index out of range panic when processing a layer that
has no entry point or command specified. Simply adds a length check before
accessing the slice.

Discovered and tested against the `apcera/gnatsd` docker image.


iaguis:
Oops, didn't know this could happen. Thanks a lot!

PR #30, opened 2015-03-17
Quote COMMAND in /bin/sh -c COMMAND

iaguis:
If we don't quote every element of the command array, spaces inside
elements will be interpreted by the shell as separators.

Fixes coreos/rocket#631

### DIFF
--- a/lib/util.go
+++ b/lib/util.go
@@ -15,6 +15,7 @@
 package docker2aci
 
 import (
+	"fmt"
 	"path"
 	"strings"
 )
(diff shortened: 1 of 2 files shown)


lib/util.go
```diff
@@ -34,3 +34,13 @@ func makeEndpointsList(headers []string) []string {
 
 	return endpoints
 }
+
+func quote(l []string) []string {
+	var quoted []string
+
+	for _, s := range l {
+		quoted = append(quoted, fmt.Sprintf("\"%s\"", s))
```
krnowak: Maybe use `"%q"`? :P Or, if you don't want to use it then use
```
`"%s"`
```
It's clearer, no escaping.

iaguis: Right, I was playing with several options and I don't know why I uploaded this one :D

krnowak:
LGTM.

iaguis:
Thanks!

PR #33, opened 2015-03-20
Generate compressed acis

### LINKED ISSUE #31: docker2aci should compress the acis it produces
It would save some disk space.

iaguis:
Until now we were generating ACIs as uncompressed tars. This compresses
the generated ACIs with gzip.

Fixes #31

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,8 @@ layers of a Docker image and squashes them into an ACI image.
 Optionally, it can generate one ACI for each layer setting the correct
 dependencies.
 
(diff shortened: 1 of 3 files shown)


iaguis:
I don't think adding the option to generate uncompressed ACIs is worth it but I can be convinced :)

jonboulle:
This LGTM, but could you add a note to the documentation and/or CLI help mentioning that this is the default behaviour? thanks

iaguis:
Will do

PR #36, opened 2015-03-26
Handle whiteout files

iaguis:
We were ignoring Docker's whiteout files so it's time to handle them.

To do that, we keep a white list of files (PathWhitelist). As we go
through the layers, we remove any files maked as whiteouts from the list
so so that, at any time, each layer ACI has the list of its desired
files and the ones of its parent layers.

We change order we process the images to start from the base image so
that we can generate the correct white lists. Then we reverse the image
list because acirenderer expects images starting from the upper layer.

### DIFF
--- a/lib/docker2aci.go
+++ b/lib/docker2aci.go
@@ -89,8 +89,11 @@ func Convert(dockerURL string, squash bool, outputDir string) ([]string, error) 
 
 	var images acirenderer.Images
 	var aciLayerPaths []string
-	for i, layerID := range ancestry {
-		aciPath, manifest, err := buildACI(layerID, repoData, parsedURL, layersOutputDir)
+	var curPwl []string
+	for i := len(ancestry) - 1; i >= 0; i-- {
+		layerID := ancestry[i]
+
+		aciPath, manifest, err := buildACI(layerID, repoData, parsedURL, layersOutputDir, curPwl)
 		if err != nil {
 			return nil, fmt.Errorf("error building layer: %v\n", err)
 		}
@@ -102,8 +105,11 @@ func Convert(dockerURL string, squash bool, outputDir string) ([]string, error) 
 
 		images = append(images, acirenderer.Image{Im: manifest, Key: key, Level: uint16(i)})
 		aciLayerPaths = append(aciLayerPaths, aciPath)
+		curPwl = manifest.PathWhitelist
 	}
 
+	// acirenderer expects images in order from upper to base layer
+	images = reverseImages(images)
 	if squash {
 		squashedImagePath, err := SquashLayers(images, conversionStore, *parsedURL, outputDir)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


jonboulle:
@iaguis thanks for tackling this. Is there any particular reason you went with the `pathWhitelist` approach as opposed to e.g. simply omitting the actual files when constructing the rootfs? I'm a bit ambivalent about this approach since we potentially have to create a giant `pathWhiteList` just to remove a single file. Thoughts?

iaguis:
You need to use `pathWhitelist` if you're generating the layers separately as ACIs. Now, for a squashed image we could do like you say but by using `pathWhitelist` I can let acirenderer do the hard work of finding which files should I get from which layer.

What we can do is remove `pathWhitelist` from the squashed manifest. Once it's squashed, it was already taken into account by acirenderer and it's redundant. Would that be enough?

lib/docker2aci.go
```diff
@@ -552,15 +557,36 @@ func writeACI(layer io.ReadSeeker, manifest schema.ImageManifest, output string)
...
-	return nil
+	manifest.PathWhitelist = newPwl
+	if err := addMinimalACIStructure(trw, manifest); err != nil {
+		return nil, fmt.Errorf("error writing rootfs entry: %v", err)
+	}
+
+	return &manifest, nil
+}
+
+func substractWhiteouts(pathWhitelist []string, whiteouts []string) []string {
```
jonboulle: subtract

iaguis: Damn, my eternal mistake in English...

lib/docker2aci.go
```diff
@@ -528,17 +534,16 @@ func writeACI(layer io.ReadSeeker, manifest schema.ImageManifest, output string)
...
+	var whiteouts []string
 	convWalker := func(t *tarball.TarFile) error {
 		name := t.Name()
 		if name == "./" {
 			return nil
 		}
 		t.Header.Name = path.Join("rootfs", name)
+		absolutePath := strings.TrimPrefix(t.Header.Name, "rootfs")
 		if strings.Contains(t.Header.Name, "/.wh.") {
+			whiteouts = append(whiteouts, strings.Replace(absolutePath, ".wh.", "", -1))
```
jonboulle: I don't know the specific legalities of filenames here, but wouldn't `1` be safer than `-1`? (e.g. `.wh.myfile.wh.ich.has.a.silly.name`)

jonboulle:
One question, LGTM aside from that

lib/docker2aci.go
iaguis: You're right

iaguis:
Thanks!

jaybuff:
+1 iaguis, thanks for doing this!

jonboulle:
@iaguis thank _you_!

PR #37, opened 2015-03-27
Iaguis/convert local images

### LINKED ISSUE #32: docker2aci should work with local docker images
Currently docker2aci can only take a docker image from a registry (through the network).

After building a Docker image locally from a Dockerfile, it is inconvenient to have to upload it to a registry before being able to convert it to an aci with docker2aci.

Instead, it would be great if docker2aci could work on local files produced by docker-save.

iaguis:
Adds support for converting Docker images generated by "docker save".
Since this file can include several images and tags, we convert whatever
image is there if there's only one or we politely ask the user to select
one image if there're several.

This also changes docker2aci's interface. We now require the prefix
"docker://" to get images from a Docker registry and assume that if it's
not present the user means a local file.

Depends on #36 
Fixes #32


jonboulle:
> This also changes docker2aci's interface. We now require the prefix
> "docker://" to get images from a Docker registry and assume that if it's
> not present the user means a local file.

I'm 50/50 on this (vs. having a `--from-file` flag or so). Is your thinking that we should have the same syntax as Rocket? Or that converting from local images will be the more common use case?

lib/common/common.go
```diff
@@ -0,0 +1,350 @@
...
+		return nil
+	}
+	reader, err := aci.NewCompressedTarReader(layer)
+	if err == nil {
+		// Write files in rootfs/
+		if err := tarball.Walk(*reader, convWalker); err != nil {
+			return nil, err
+		}
+	} else {
+		// ignore errors
```
jonboulle: eh?

lib/common/common.go
```diff
@@ -0,0 +1,350 @@
...
+	}
+	newPwl := subtractWhiteouts(curPwl, whiteouts)
+
+	manifest.PathWhitelist = newPwl
+	if err := addMinimalACIStructure(trw, manifest); err != nil {
+		return nil, fmt.Errorf("error writing rootfs entry: %v", err)
```
jonboulle: This is slightly misleading since this also writes the manifest. Although I am wondering if it would be better to write the rootfs dir before doing the convWalker anyway..

lib/docker2aci.go
```diff
@@ -55,25 +55,38 @@ const (
...
-	if err != nil {
-		return nil, fmt.Errorf("error getting ImageID from tag %s: %v\n", parsedURL.Tag, err)
-	}
+	localBackend := local.NewLocalBackend(file)
+	return convertReal(localBackend, dockerURL, squash, outputDir)
+}
 
-	ancestry, err := getAncestry(appImageID, repoData.Endpoints[0], repoData)
+func convertReal(backend Docker2ACIBackend, dockerURL string, squash bool, outputDir string) ([]string, error) {
+	util.Debug("Getting image info from the registry...")
```
jonboulle: needs update, might no longer be registry

lib/backend/local/local.go
```diff
@@ -0,0 +1,331 @@
...
+	_, err := file.Seek(0, 0)
+	if err != nil {
+		return "", fmt.Errorf("error seeking file: %v", err)
+	}
+
+	jsonPath := filepath.Join(imgID, "json")
+	parentWalker := func(t *tarball.TarFile) error {
+		cleanName := filepath.Clean(t.Name())
+
+		if cleanName == jsonPath {
```
jonboulle: collapse with line 306

lib/backend/local/local.go
```diff
@@ -0,0 +1,331 @@
...
+	"strings"
+
+	"github.com/appc/docker2aci/lib/common"
+	"github.com/appc/docker2aci/lib/types"
+	"github.com/appc/docker2aci/lib/util"
+	"github.com/appc/docker2aci/tarball"
+	"github.com/appc/spec/schema"
+)
+
+type LocalBackend struct {
```
jonboulle: To be a bit more specific I suggest "FileBackend" vs. "RepositoryBackend"

lib/backend/local/local.go
```diff
@@ -0,0 +1,331 @@
...
+
+		if cleanName == "repositories" {
+			repob, err := ioutil.ReadAll(t.TarStream)
+			if err != nil {
+				return fmt.Errorf("error reading repositories file: %v", err)
+			}
+
+			var repositories apps
+			if err := json.Unmarshal(repob, &repositories); err != nil {
+				return fmt.Errorf("error Unmarshaling repositories file")
```
jonboulle: unmarshaling

lib/backend/local/local.go
```diff
@@ -0,0 +1,331 @@
...
+		fmt.Errorf("error seeking file: %v", err)
+	}
+
+	layerTarPath := path.Join(layerID, "layer.tar")
+
+	var layerFile *os.File
+	fileWalker := func(t *tarball.TarFile) error {
+		cleanName := filepath.Clean(t.Name())
+
+		if cleanName == layerTarPath {
```
jonboulle: collapse with 250

lib/backend/local/local.go
```diff
@@ -0,0 +1,331 @@
...
+	if err != nil {
+		return "", nil, fmt.Errorf("error seeking file: %v", err)
+	}
+
+	var imageID string
+	var appName string
+	reposWalker := func(t *tarball.TarFile) error {
+		cleanName := filepath.Clean(t.Name())
+
+		if cleanName == "repositories" {
```
jonboulle: collapse with 93

lib/backend/local/local.go
```diff
@@ -0,0 +1,331 @@
...
+			}
+
+			if dockerURL == nil {
+				if len(repositories) > 1 {
+					var appNames []string
+					for key, _ := range repositories {
+						appNames = append(appNames, key)
+					}
+					return fmt.Errorf("several images found, use option --image with one of:\n\n%s", strings.Join(appNames, "\n"))
+				} else {
```
jonboulle: maybe check for len == 0 just in case?

lib/backend/local/local.go
```diff
@@ -0,0 +1,331 @@
...
+			} else {
+				appName = dockerURL.ImageName
+			}
+
+			tag := "latest"
+			if dockerURL != nil {
+				tag = dockerURL.Tag
+			}
+
+			if _, ok := repositories[appName]; !ok {
```
jonboulle: be optimistic and assign the value to something so you don't have to keep dereferencing it again below

lib/common/common.go
```diff
@@ -0,0 +1,350 @@
...
+	"github.com/appc/docker2aci/lib/types"
+	"github.com/appc/docker2aci/lib/util"
+	"github.com/appc/spec/aci"
+	"github.com/appc/spec/schema"
+	appctypes "github.com/appc/spec/schema/types"
+)
+
+const (
+	defaultTag    = "latest"
+	schemaVersion = "0.1.1"
```
jonboulle: this can be bumped right?

lib/common/common.go
```diff
@@ -0,0 +1,350 @@
...
+	aciPath += ".aci"
+
+	aciPath = path.Join(outputDir, aciPath)
+	manifest, err = writeACI(layerFile, *manifest, curPwl, aciPath)
+	if err != nil {
+		return "", nil, fmt.Errorf("error writing ACI: %v", err)
+	}
+
+	if err := ValidateACI(aciPath); err != nil {
+		return "", nil, fmt.Errorf("invalid aci generated: %v", err)
```
jonboulle: nit, inconsistent aci vs ACI in messages

lib/common/common.go
```diff
@@ -0,0 +1,350 @@
...
+	if err != nil {
+		return nil, err
+	}
+	name, err := appctypes.NewACName(appURL)
+	if err != nil {
+		return nil, err
+	}
+	genManifest.Name = *name
+
+	acVersion, _ := appctypes.NewSemVer(schemaVersion)
```
jonboulle: might as well check the error here + panic, just in case

lib/common/common.go
```diff
@@ -0,0 +1,350 @@
...
+		parentAppNameString, err := appctypes.SanitizeACName(parentAppNameString)
+		if err != nil {
+			return nil, err
+		}
+		parentAppName, err := appctypes.NewACName(parentAppNameString)
+		if err != nil {
+			return nil, err
+		}
+
+		dependencies = append(dependencies, appctypes.Dependency{App: *parentAppName, Labels: parentLabels})
```
jonboulle: something is weird here, maybe got lost in a historical refactoring, but afaict `var dependencies` can be scrapped and you can just assign directly to `genManifest.Dependencies = appctypes.Dependencies{appctypes.Dependency{App: *parentAppName, Labels: parentLabels})`

lib/common/common.go
```diff
@@ -0,0 +1,350 @@
...
+	name, err := appctypes.NewACName(appURL)
+	if err != nil {
+		return nil, err
+	}
+	genManifest.Name = *name
+
+	acVersion, _ := appctypes.NewSemVer(schemaVersion)
+	genManifest.ACVersion = *acVersion
+
+	genManifest.ACKind = appctypes.ACKind("ImageManifest")
```
jonboulle: schema.ImageManifestKind

lib/common/common.go
```diff
@@ -0,0 +1,350 @@
...
+
+	var labels appctypes.Labels
+	var parentLabels appctypes.Labels
+
+	layer, _ := appctypes.NewACName("layer")
```
jonboulle: Maybe we should have a MustACName(), would obviate a lot of these lines

lib/common/common.go
```diff
@@ -0,0 +1,350 @@
...
+	return nil
+}
+
+func WriteRootfsDir(tarWriter *tar.Writer) error {
+	hdr := getGenericTarHeader()
+	hdr.Name = "rootfs"
+	hdr.Mode = 0755
+	hdr.Size = int64(0)
+	hdr.Typeflag = tar.TypeDir
+
```
jonboulle: return tarWriter.WriteHeader(hdr)

main.go
```diff
@@ -17,24 +17,52 @@ package main
...
-func runDocker2ACI(arg string, flagNoSquash bool) error {
+func runDocker2ACI(arg string, flagNoSquash bool, flagImage string, flagDebug bool) error {
+	if flagDebug {
+		util.InitDebug()
+	}
 	squash := !flagNoSquash
 
-	aciLayerPaths, err := docker2aci.Convert(arg, squash, ".")
+	var aciLayerPaths []string
+	var err error
```
jonboulle: redundant

main.go
```diff
@@ -17,24 +17,52 @@ package main
...
+		return fmt.Errorf("error parsing argument: %v", err)
+	}
+	if u.Scheme == "docker" {
+		if flagImage != "" {
+			return fmt.Errorf("flag --image works only with files.")
+		}
+		registryURL := strings.TrimPrefix(arg, "docker://")
+
+		aciLayerPaths, err = docker2aci.Convert(registryURL, squash, ".")
+		if err != nil {
```
jonboulle: split this check below the if/else

main.go
```diff
@@ -17,24 +17,52 @@ package main
...
 )
 
 const (
 	rocketDir = "/var/lib/rkt"
 )
 
-var flagNoSquash = flag.Bool("nosquash", false, "Don't Squash layers and output every layer as ACI")
+var (
+	flagNoSquash = flag.Bool("nosquash", false, "Don't Squash layers and output every layer as ACI")
+	flagImage    = flag.String("image", "", "When converting a local file, it selects a particuler image to convert. Format: IMAGE_NAME[:TAG]")
```
jonboulle: particular

main.go
```diff
@@ -17,24 +17,52 @@ package main
...
+	"github.com/appc/docker2aci/lib/util"
 )
 
 const (
 	rocketDir = "/var/lib/rkt"
 )
 
-var flagNoSquash = flag.Bool("nosquash", false, "Don't Squash layers and output every layer as ACI")
+var (
+	flagNoSquash = flag.Bool("nosquash", false, "Don't Squash layers and output every layer as ACI")
```
jonboulle: s#Squash#squash# please

lib/common/common.go
iaguis: When a layer's rootfs is empty it's not a valid tar file in files generated by docker save so NewCompressedTarReader will fail. In that case, ignoring the error will result in an ACI with a manifest and an empty rootfs directory, which is what we want.

I can add a better comment.

lib/common/common.go
iaguis: Yep

lib/common/common.go
iaguis: OK. I'll remove addMinimalACIStructure(), write the rootfs dir before convWalker and the manifest here.

lib/common/common.go
iaguis: appc/spec#274

iaguis:
>> This also changes docker2aci's interface. We now require the prefix
>> "docker://" to get images from a Docker registry and assume that if it's
>> not present the user means a local file.

> I'm 50/50 on this (vs. having a --from-file flag or so). Is your thinking that we should have the same syntax as Rocket? Or that converting from local images will be the more common use case?

Yeah, I'm also on the fence about that. I implemented it this way to keep it simple: adding a --from-file is less convenient for what I think would be a pretty common use case. Since we're already using this syntax in Rocket it seemed a pretty good choice. I don't really know if converting from local images will be the most common use case, probably not.

We also have to think whether we want rocket to support converting files generated by `docker save` and how to expose it...

iaguis:
Thanks a lot for the review!

jonboulle:
> We also have to think whether we want rocket to support converting files generated by docker save and how to expose it...

I don't think this is necessary, if people really have exported Docker images around it seems reasonable to expect them to use `docker2aci` (it's just one more command). My feeling is that the Rocket auto-Docker stuff is mostly useful for referencing images from repositories.

jonboulle:
> Yeah, I'm also on the fence about that. I implemented it this way to keep it simple: adding a --from-file is less convenient for what I think would be a pretty common use case. Since we're already using this syntax in Rocket it seemed a pretty good choice. I don't really know if converting from local images will be the most common use case, probably not.

@philips any feelings on this?

philips:
@jonboulle @iaguis Can you provide a CLI example. I am having a hard time following.

jonboulle:
@philips 
```
docker2aci quay.io/coreos/etcd
docker2aci --from-file ./etcd
```
vs.
```
docker2aci docker://quay.io/coreos/etcd
docker2aci ./etcd
```

philips:
I see. The other option is to just always make it a url like file://etcd

jonboulle:
Sure, I consider ./ a minor optimisation over that

philips:
Yea, we just need to be explicit. The current state in rkt bothers and we need to fix it because the rules are ambiguous and can lead to mistakes or confusion.

This is a stupid example:

```
cd $HOME/src
./github.com/coreos/rocket/bin/rkt run github.com/coreos/etcd
github.com/coreos/etcd: error reading image header: read github.com/coreos/etcd: is a directory
```

jonboulle:
@philips agreed, it just hasn't bothered me personally enough yet to
prioritise it. Do you want to file another readme-driven-development issue
on coreos/rkt?

philips:
@jonboulle Yes: https://github.com/coreos/rkt/issues/715

jonboulle:
danke!

jonboulle:
I propose we land this as-is and can revisit the exact UX after we sort it out in rkt, @philips that sound reasonable to you?

philips:
@jonboulle Sure, go for it.

PR #38, opened 2015-03-31
Compress layers only when not squashing

### LINKED ISSUE #34: Converting big images is slow
I did some benchmarks with big images after #33 and the conversion is around 40% slower. This is so because we're compressing each layer and decompressing it when we squash to compress the squashed image again.

I think generating one ACI for each layer is a rather uncommon use case so compressing only the squashed image would be acceptable.

We could also consider using https://github.com/klauspost/pgzip to make it even faster.

iaguis:
Depends on #37
Fixes #34


jonboulle:
c12df2e LGTM but I guess we gotta get #37 in first..

iaguis:
Fixed the nits and #37 is in so I'm merging this

PR #43, opened 2015-04-14
Convert Docker Volumes to mountPoints

iaguis:
A Docker image can specify Volumes which are akin to mountPoints in the
appc spec. This commit converts Volumes to mountPoints allowing the user
to mount volumes with rkt's --volume option.

Since Docker doesn't name Volumes but the appc spec requires a name, we
generate a name by appending the path to "volume-". That is, if a Volume
has "/var/tmp" as path, the resulting mountPoint name will be
"volume-/var/tmp". A mountPoint name must be a valid ACName so any
invalid characters will be replaced by a "-".

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,17 @@ the correct dependencies.
 
 All ACIs generated are compressed with gzip.
 
-[aci]: https://github.com/appc/spec/blob/master/SPEC.md#app-container-image
+
+## Volumes
+
+Docker Volumes get converted to mountPoints in the
+[Image Manifest Schema][imageschema]. Since mountPoints need a name and Docker
+Volumes don't, docker2aci generates a name by appending the path to `volume-`.
+That is, if a Volume has `/var/tmp` as path, the resulting mountPoint name will
+be `volume-/var/tmp`.
+
+When the docker2aci CLI binary converts a Docker Volume to a mountPoint it will
+print its name, path and whether it is read-only or not.
 
 ## CLI examples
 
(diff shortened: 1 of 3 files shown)


jonboulle:
Could you document this, please?

iaguis:
Done.

rimusz:
@iaguis :+1:

lib/common/common.go
```diff
@@ -161,6 +161,20 @@ func GenerateManifest(layerData types.DockerImageData, dockerURL *types.ParsedDo
 			}
 			genManifest.App = app
 		}
+
+		for p, _ := range dockerConfig.Volumes {
+			n := filepath.Join("volume-", p)
+			sn, err := appctypes.SanitizeACName(n)
```
jonboulle: Hrm maybe a corner case but don't we need to deal with duplicates? e.g. `/data#`, `/data$`

jonboulle:
LGTM modulo question about duplicates.

lib/common/common.go
iaguis: Yes we do, thanks!

lib/common/common.go
```diff
@@ -177,6 +183,36 @@ func GenerateManifest(layerData types.DockerImageData, dockerURL *types.ParsedDo
 	return genManifest, nil
 }
 
+func convertVolumesToMPs(dockerVolumes map[string]struct{}) ([]appctypes.MountPoint, error) {
+	mps := []appctypes.MountPoint{}
+	dup := make(map[string]int)
+
+	for p, _ := range dockerVolumes {
```
krnowak: No need for the underline, you can simply type `p := range dockerVolumes`

lib/common/common.go
```diff
@@ -177,6 +183,36 @@ func GenerateManifest(layerData types.DockerImageData, dockerURL *types.ParsedDo
...
+		n := filepath.Join("volume-", p)
+		sn, err := appctypes.SanitizeACName(n)
+		if err != nil {
+			return nil, err
+		}
+
+		// check for duplicate names
+		if i, ok := dup[sn]; ok {
+			dup[sn] = i + 1
+			sn = fmt.Sprintf("%s%s%d", sn, "-", i)
```
krnowak: Maybe just use `fmt.Sprintf("%s-%d", sn, i)`?

lib/common/common.go
iaguis: Right

lib/common/common.go
iaguis: Yeah, that makes it more clear

lib/common/common.go
```diff
@@ -177,6 +183,36 @@ func GenerateManifest(layerData types.DockerImageData, dockerURL *types.ParsedDo
...
+		} else {
+			dup[sn] = 1
+		}
+
+		mp := appctypes.MountPoint{
```
krnowak: The Readonly member is not set, so it is false by default? Does it match behaviour in docker?

iaguis: http://docs.docker.com/userguide/dockervolumes/#mount-a-host-directory-as-a-data-volume

"Docker defaults to a read-write volume but we can also mount a directory read-only."

main.go
```diff
@@ -65,6 +68,12 @@ func runDocker2ACI(arg string, flagNoSquash bool, flagImage string, flagDebug bo
 		return fmt.Errorf("conversion error: %v", err)
 	}
 
+	if squash {
```
krnowak: What about printing converted volumes in non-squashing scenario?

lib/common/common.go
krnowak: Alright, thanks.

main.go
iaguis: I can add it. It looks a bit ugly because there will be a lot of repeated volumes:

```
Converted volumes in "postgres-26bdc4facea7aad8cfdf3da846e3767bfa027fc0b3c621501f8d7a2413fd24e2-latest-linux-amd64.aci":
        name: "volume-/var/lib/postgresql/data", path: "/var/lib/postgresql/data", readOnly: false

Converted volumes in "postgres-3508d36eab9991998d587938711fd19286205eb1a7e994fc4df804ab08e14013-latest-linux-amd64.aci":
        name: "volume-/var/lib/postgresql/data", path: "/var/lib/postgresql/data", readOnly: false

Converted volumes in "postgres-86cff6f79a0869259cd911587f61282d9c7a5afc87ca425290cfc781cfaa83bd-latest-linux-amd64.aci":
        name: "volume-/var/lib/postgresql/data", path: "/var/lib/postgresql/data", readOnly: false

Converted volumes in "postgres-9c5489b55573bdc2e214f35ed102543845c0de6ca5df10eb5c037ee93a126002-latest-linux-amd64.aci":
        name: "volume-/var/lib/postgresql/data", path: "/var/lib/postgresql/data", readOnly: false

Converted volumes in "postgres-dcce0da1c31105323a840ded6e79f39b7338907f5c2477109b68692e7acb75be-latest-linux-amd64.aci":
        name: "volume-/var/lib/postgresql/data", path: "/var/lib/postgresql/data", readOnly: false
```

krnowak: Hm, then maybe deduping it beforehand would be useful?

iaguis: OK, I dedup the mountpoints before printing them

iaguis: And I hate go for its lack of generics

README.md
```diff
@@ -51,3 +61,34 @@ Extracting layer: 2103b00b3fdf1d26a86aded36ae73c1c425def0f779a6e69073b3b77377df3
...
+Downloading layer: 607eb184411289059954fec63a321ef6408958d52efbf97966fb134b7cbeccca
+Downloading layer: a8beffce6b07225b9dfc2458c049e901cb748c1b58a6130a493fba2fd35b6b9e
+Downloading layer: b23ebc9e057634e7b402bb63ca835483796ecf60efd2e94be8eb45d0756851d0
+Downloading layer: c817c8c769d0120041f428a89ae270d4b616f8eed760b1cc96b0efaecd5e78d8
+Downloading layer: 19682c2cca49564c691ac06052de49f61d5bdf16c9203cf83b123903a6f052eb
+Downloading layer: 4a418129851356d0f07dc8be4455c3f8a15d11ac4bb3410eb799a647da6832ba
+Downloading layer: d47fc7449244382dc6121e68100a2c677f4964b495c5bdbbe2975be21777151e
+Downloading layer: f2fb89b0a711a7178528c7785d247ba3572924353b0d5e23e9b28f0518253b22
+
+Converted volumes in "redis-latest.aci":
```
krnowak: This actually is not printed. :P

iaguis: This is still a WIP :grimacing:

iaguis:
Updated.

main.go
```diff
@@ -73,6 +80,41 @@ func runDocker2ACI(arg string, flagNoSquash bool, flagImage string, flagDebug bo
 	return nil
 }
 
+func printConvertedVolumes(aciPaths []string) error {
```
krnowak: I would pass just a string instead of an array. Or at least document why we are using only last element here.

krnowak:
LGTM.

iaguis:
:smiley:

PR #44, opened 2015-04-14
Use MustACName when we know creation won't fail

iaguis:
If we use known-to-be-valid strings when creating new ACNames or if we
previously called SanitizeACName we know the creation won't fail.


jonboulle:
:+1:, thanks for following this up.

One (general) nit: could you please title your commits with the `<subsystem>: <what changed>` format we try to use elsewhere?
(should probably add a CONTRIBUTING guide to this repo - https://github.com/appc/spec/blob/master/CONTRIBUTING.md#format-of-the-commit-message  )

Other than that, lgtm

iaguis:
Sure

Issue #50, opened 2015-04-28
support configurable tmpdir


jonboulle:
As a library, docker2aci is used by tools like rkt to transparently retrieve + convert Docker images for users. In some cases these tools have strict requirements about their filesystem writes. Currently docker2aci makes heavy use of temporary files when it retrieves images. To integrate more cleanly with tools like rkt, it should support being passed a configuration value controlling where it writes these files.

PR #51, opened 2015-04-28
lib: add the tmpDir parameter

alban:
Fixes: https://github.com/appc/docker2aci/issues/50

### DIFF
--- a/lib/docker2aci.go
+++ b/lib/docker2aci.go
@@ -53,11 +53,13 @@ type Docker2ACIBackend interface {
 // If the squash flag is true, it squashes all the layers in one file and
 // places this file in outputDir; if it is false, it places every layer in its
 // own ACI in outputDir.
+// It will use the temporary directory specified by tmpDir, or the default
+// temporary directory if tmpDir is "".
 // username and password can be passed if the image needs authentication.
 // It returns the list of generated ACI paths.
-func Convert(dockerURL string, squash bool, outputDir string, username string, password string) ([]string, error) {
+func Convert(dockerURL string, squash bool, outputDir string, tmpDir string, username string, password string) ([]string, error) {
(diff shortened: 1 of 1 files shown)


lib/docker2aci.go
```diff
@@ -103,7 +107,7 @@ func convertReal(backend Docker2ACIBackend, dockerURL string, squash bool, outpu
 	}
 
 	layersOutputDir := outputDir
-	if squash {
+	if squash && tmpDir == "" {
```
jonboulle: hmm? Shouldn't you be passing this into ioutil.TempDir below?

alban: You're right of course.

lib/docker2aci.go
```diff
@@ -53,11 +53,13 @@ type Docker2ACIBackend interface {
 // If the squash flag is true, it squashes all the layers in one file and
 // places this file in outputDir; if it is false, it places every layer in its
 // own ACI in outputDir.
+// It will use the temporary directory specified by tmpDir, or the default
+// temporary directory if tmpDir is "".
 // username and password can be passed if the image needs authentication.
 // It returns the list of generated ACI paths.
-func Convert(dockerURL string, squash bool, outputDir string, username string, password string) ([]string, error) {
+func Convert(dockerURL string, squash bool, outputDir string, tmpDir string, username string, password string) ([]string, error) {
```
krnowak: Function with six parameters, ugh. This is getting out of hand.

Feel free to ignore this comment, of course.

jonboulle: I agree :/

jonboulle:
LGTM unless you would like to clean up the args.

alban:
Patch rebased on master and merged.

PR #53, opened 2015-04-30
lib/common: include annotations from docker image

vbatts:
* author
* created
* comment

/cc @philips

### DIFF
--- a/lib/common/common.go
+++ b/lib/common/common.go
@@ -5,12 +5,13 @@ import (
 	"compress/gzip"
 	"encoding/json"
 	"fmt"
-	"github.com/appc/docker2aci/tarball"
 	"io"
 	"os"
 	"path"
 	"path/filepath"
 	"time"
+
+	"github.com/appc/docker2aci/tarball"
 
 	"strings"
 
@@ -120,8 +121,11 @@ func GenerateManifest(layerData types.DockerImageData, dockerURL *types.ParsedDo
 
 	genManifest.ACKind = appctypes.ACKind(schema.ImageManifestKind)
 
-	var labels appctypes.Labels
-	var parentLabels appctypes.Labels
+	var (
+		labels       appctypes.Labels
+		parentLabels appctypes.Labels
+		annotations  appctypes.Annotations
+	)
 
 	layer := appctypes.MustACName("layer")
 	labels = append(labels, appctypes.Label{Name: *layer, Value: layerData.ID})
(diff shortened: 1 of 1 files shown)


philips:
Created and authors make sense to me. Comment seems pretty nebulous. Can we describe what a comment is and put it in the spec?

vbatts:
I reckon it would be just as well `description`, because that is how it's used. I can drop it if you like, but `annotations` seem arbitrary. We could just as well, include information about where this image was converted from.

philips:
@vbatts the annotations are arbitrary but if we are using the top-level namespace I want to write-down what people can expect to find in each field. Otherwise the field is sort of useless. By namespace it could be docker.io/comment.

vbatts:
... or  `io.docker.$imagename.comment` ? (ala https://github.com/projectatomic/ContainerApplicationGenericLabels bullet 3)

philips:
@vbatts Why is the imagename necessary? The annotations are per image?

vbatts:
if the namespace is where it is from, then that information is per image or ID.

philips:
@vbatts I don't quite follow. Could you provide an example fo what you mean by namespace?

vbatts:
I was scoping to the source (registry and image name) then field.

I'm not terribly opposed here, one way or the other. If these annotation
field names are essential to pin, then there might need to be more
strictness around them. :-)

philips:
I don't mind adding stuff I just want to write down what it is used for so when some dashboard, cli tool, etc finds it they can reasonably expect what size/shape it will be when they print it out.

Scoping on the registry/image name doesn't make sense to me since this is the metadata for the image. So the name of the image should be enough of a scope.

vbatts:
@philips should all three values be prefixed? `docker.author`, `docker.created`, and `docker.comment`

philips:
@vbatts author and created are in the appc spec today and it seems in-line with what is expected in docker so I am good on that. If we want to add a comment we just need to decide on what the comment is, document it and then use it :)

See annotations immediately above this section: https://github.com/appc/spec/blob/master/SPEC.md#dependency-matching

vbatts:
i've pushed only `comment` having a namespace prefix for now. That field _could_ blend to `description` for annotations, but does not have docs to back that up. Likely just a loose field.

philips:
@vbatts we have been using ACName which is defined as a dns name separated by slashes. So:

docker.io/comment

Otherwise LGTM!

vbatts:
ah, i just chose not the dns style, since the image may have been converted
from another source than the hub, but came from a docker format.

philips:
@vbatts hrm, appc.io/docker/comment, docker/comment? Thoughts @jonboulle ?

jonboulle:
oh boy if only there was a standard :trollface: 
No strong opinion sorry. appc.io/docker/comment probably slightly better for now.

vbatts:
ah. i didn't see your reply til now. i pushed a change to `docker/comment`.
I would lead the `appc.io/` prefix off here.

vbatts:
PTAL

jonboulle:
Seems fine for now, we can revisit if a greater consensus emerges around
this. Lgtm!
Am 01.05.2015 19:12 schrieb "Vincent Batts" <notifications@github.com>:

philips:
LGTM, as there is no MAINTAINERS file I am going to merge.

@iaguis @alban @jonboulle @krnowak can someone please add a maintainers file?

PR #55, opened 2015-04-30
all: use godep to pin dependencies

### LINKED ISSUE #49: vendor dependencies with godep
see https://github.com/tools/godep

mischief:
fixes #49


iaguis:
Thanks, LGTM

mischief:
@iaguis i apologize for the code churn but i think you will have to revert this. i made this pr with the intention of locking dependencies for the docker2aci program, without realizing that it is also used as a library by other go programs. https://github.com/tools/godep/issues/109#issuecomment-52143649 describes in greater detail why libraries should not use godep.

the only reasonable way i can see to keep godep out of docker2aci/lib *and* pin dependencies for the docker2aci program (main.go) is to move main.go into a different repo and use godep there, but i'm not sure if that is desireable.

:disappointed:

iaguis:
OK, let's revert it for now...

PR #60, opened 2015-05-20
*.*: Replace stdout with stderr for human consumable messages.

yifan-gu:


lib/backend/repository/repository.go
```diff
@@ -302,7 +302,7 @@ func getLayer(imgID, registry string, repoData *RepoData, imgSize int64, tmpDir
 	progressReader := &ioprogress.Reader{
 		Reader:       res.Body,
 		Size:         imgSize,
-		DrawFunc:     ioprogress.DrawTerminalf(os.Stdout, fmtfunc),
+		DrawFunc:     ioprogress.DrawTerminalf(os.Stderr, fmtfunc),
```
jonboulle: This is a lib and really shouldn't be printing to arbitrary streams - it should probably accept an optional io.Writer that users of the lib can inject. Would you mind filing a followup issue?

jonboulle:
LGTM for now

lib/backend/repository/repository.go
yifan-gu: ? Is providing the stream via DrawFunc not enought?

jonboulle: What I mean is that users of docker2aci should not have it printing to stderr unless they request it

yifan-gu: ok. I opened an issue for this and will follow up.

jonboulle:
@yifan-gu thank you!

PR #64, opened 2015-05-23
Fixed the build of docker2aci

kontsevoy:
docker2aci.ConvertXXX() has a new parameter which wasn't added to
main.go

Setting to "/tmp".


iaguis:
Good catch. Can you use os.TempDir() instead of hardcoding "/tmp"?

kontsevoy:
Oh sweet. TIL. Done.

iaguis:
Also, can you remove the binary file from the commit? It shouldn't be in the repo. Thanks!

kontsevoy:
done

PR #65, opened 2015-05-26
lib: add the tmpDir parameter (again)

### LINKED ISSUE #50: support configurable tmpdir

alban:
First, this revert a patch so that docker2aci compiles against appc/spec v0.5.2 rather than just git-master. It will be useful for rkt, see https://github.com/coreos/rkt/pull/929#issuecomment-104787060.

Then, add the tmpDir parameter to fix #50. It was partially done in #51 but not completely.

### DIFF
--- a/lib/backend/file/file.go
+++ b/lib/backend/file/file.go
@@ -44,8 +44,8 @@ func (lb *FileBackend) GetImageInfo(dockerURL string) ([]string, *types.ParsedDo
 	return ancestry, parsedDockerURL, nil
 }
 
-func (lb *FileBackend) BuildACI(layerID string, dockerURL *types.ParsedDockerURL, outputDir string, curPwl []string, compress bool) (string, *schema.ImageManifest, error) {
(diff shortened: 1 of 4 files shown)


iaguis:
This breaks the build but we need a ref with the tmpdir parameter working and working against appc/spec v0.5.2. Please add a revert of [Revert "common: rename App to ImageName"](https://github.com/endocode/docker2aci/commit/0345b92435e6cf402d87a9e55d6248402bbe537e) so the build doesn't break but you still can vendor the intermediary commit in rkt.

After that, LGTM.

iaguis:
I like reverting reverts. LGTM.

PR #66, opened 2015-06-01
common: convert docker ports

iaguis:
Convert docker ExposedPorts to appc's []types.Port.

The Name field is taken from docker's port representation (e.g. "8080"
or "8080/udp") and port number and protocol are converted in the obvious
way. If the docker image doesn't specify a protocol, we assume tcp[1]

Also, support deprecated docker PortSpecs.

[1]: https://github.com/docker/docker/blob/master/image/spec/v1.md

### DIFF
--- a/lib/common/common.go
+++ b/lib/common/common.go
@@ -9,14 +9,13 @@ import (
 	"os"
 	"path"
 	"path/filepath"
-	"time"
-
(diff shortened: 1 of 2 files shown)


jonboulle:
LGTM, thanks.

PR #67, opened 2015-06-02
common: revert rename App to ImageName and revert the revert

iaguis:
This PR only has one purpose: to get a commit in docker2aci master we can reference from rkt that is compatible with appc/spec v0.5.2 and includes the port-conversion feature.


jonboulle:
:-(
sorry @iaguis .
LGTM

iaguis:
No worries ;)

PR #68, opened 2015-06-10
Fix build by replacing ACName with ACIdentifier as in the spec

slallema:
Hi,
docker2aci wasn't building anymore so I tried to fix this by changing ACName to ACIdentifier where needed.
I think this was broken since https://github.com/appc/spec/pull/398 has been merged.

### DIFF
--- a/lib/common/common.go
+++ b/lib/common/common.go
@@ -105,11 +105,11 @@ func GenerateManifest(layerData types.DockerImageData, dockerURL *types.ParsedDo
 		appURL = dockerURL.IndexURL + "/"
 	}
 	appURL += dockerURL.ImageName + "-" + layerData.ID
-	appURL, err := appctypes.SanitizeACName(appURL)
+	appURL, err := appctypes.SanitizeACIdentifier(appURL)
 	if err != nil {
 		return nil, err
 	}
-	name := appctypes.MustACName(appURL)
+	name := appctypes.MustACIdentifier(appURL)
 	genManifest.Name = *name
 
 	acVersion, err := appctypes.NewSemVer(schemaVersion)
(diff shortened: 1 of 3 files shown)


jonboulle:
thanks! /cc @yifan-gu @iaguis

lib/common/common.go
```diff
@@ -262,7 +262,7 @@ func convertVolumesToMPs(dockerVolumes map[string]struct{}) ([]appctypes.MountPo
 
 	for p := range dockerVolumes {
 		n := filepath.Join("volume-", p)
-		sn, err := appctypes.SanitizeACName(n)
+		sn, err := appctypes.SanitizeACIdentifier(n)
```
yifan-gu: I think Volume still uses ACName

yifan-gu:
@slallema Thanks for the fix! LGTM except for one comment above :)

lib/docker2aci.go
```diff
@@ -281,7 +281,7 @@ func mergeManifests(manifests []schema.ImageManifest) schema.ImageManifest {
 	}
 
 	// this can't fail because the old name is legal
-	nameWithoutLayerID, _ := appctypes.NewACName(stripLayerID(manifest.Name.String()))
+	nameWithoutLayerID, _ := appctypes.NewACIdentifier(stripLayerID(manifest.Name.String()))
```
iaguis: You can use MustACIdentifier and remove the comment above

iaguis:
Apart from the two comments LGTM

slallema:
Using ACName for volume name isn't working because right now the volumes names include the path : ```name: "volume-/data", path: "/data", readOnly: false``` and the spec regex for ACName has changed recently. 
Any idea on how we should handle that ?

iaguis:
So the way the PR is not it works except for Comments (there's the string `docker/comment`), also the ports conversion uses things like `80/tcp`. I propose to replace slashes with dashes, `docker-comment` and `80-tcp`.

Volume names get converted with `SanitizeACName` so they're fine, every non-valid char would be converted to a dash, the only think that needs to change is

```
-               n := filepath.Join("volume-", p)
+               n := filepath.Join("volume", p)
```

so resulting volumes would be `volume-var-log-journal` for example.

slallema:
I've done the changes, and it's working again for volumes and ports ! :) 
now the output looks like : 
```
Converted volumes:
	name: "volume-data", path: "/data", readOnly: false

Converted ports:
	name: "6379-tcp", protocol: "tcp", port: 6379, count: 1, socketActivated: false

Generated ACI(s):
redis-latest.aci
```

jonboulle:
@iaguis happy with this?

iaguis:
LGTM

PR #75, opened 2015-07-31
common: add arch label to the manifest

iaguis:
We forgot to add the arch label when converting Docker layer data to an
image manifest.


alban:
LGTM

PR #76, opened 2015-08-05
lib: properly write squashed ACI to disk

### LINKED ISSUE #73: failed to convert docker://kubernetes/redis-slave:v2
```
$ sudo /home/yifan/rkt/build-rkt-0.7.0+git/bin/rkt --debug=false --insecure-skip-verify=true fetch docker://kubernetes/redis-slave:v2
[sudo] password for yifan: 
Downloading 511136ea3c5a: [====================================] 32 B/32 B
Downloading f0dde87450ec: [====================================] 65.8 MB/65.8 MB
Downloading 76b658ecb564: [====================================] 71.5 KB/71.5 KB
Downloading 4faa69f72743: [====================================] 679 B/679 B
Downloading 2103b00b3fdf: [====================================] 32 B/32 B
Downloading 4303edc5e986: [====================================] 85.4 MB/85.4 MB
Downloading 14dfde0f38c5: [====================================] 643 B/643 B
Downloading d489fc3e869f: [====================================] 438 B/438 B
Downloading a8253f7d9cc1: [====================================] 22.7 KB/22.7 KB
Downloading 38cf7a363b14: [====================================] 32 B/32 B
Downloading d9343f29a1c4: [====================================] 32 B/32 B
Downloading 9cb3dca21a1f: [====================================] 32 B/32 B
Downloading c9eda0ec97c0: [====================================] 2.93 MB/2.93 MB
Downloading a44f8ddf5fb8: [====================================] 32 B/32 B
Downloading cf0fc1a9cd2b: [====================================] 32 B/32 B
Downloading 8567350addd7: [====================================] 32 B/32 B
Downloading dff9dfa9220d: [====================================] 32 B/32 B
Downloading 3fd8168c2f23: [====================================] 511 B/511 B
Downloading 85b1106c3ae7: [====================================] 544 B/544 B
Downloading 48a1e2b8f859: [====================================] 32 B/32 B
error converting docker image to ACI: error squashing image: error validating image: flate: corrupt input before offset 105339330
```

iaguis:
We were writing the file directly to its final destination instead of
writing it to a temporary file and then renaming it. This caused file
corruption when running docker2aci several times in parallel.

We now write the squashed ACI to a temporary file with a random name and
rename it atomically after the write is finished and the ACI validated.

Fixes #73

### DIFF
--- a/lib/docker2aci.go
+++ b/lib/docker2aci.go
@@ -23,7 +23,6 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
-	"path"
 	"path/filepath"
 	"strings"
 
@@ -154,7 +153,7 @@ func convertReal(backend Docker2ACIBackend, dockerURL string, squash bool, outpu
 
 // SquashLayers receives a list of ACI layer file names ordered from base image
 // to application image and squashes them into one ACI
-func SquashLayers(images []acirenderer.Image, aciRegistry acirenderer.ACIRegistry, parsedDockerURL types.ParsedDockerURL, outputDir string) (string, error) {
+func SquashLayers(images []acirenderer.Image, aciRegistry acirenderer.ACIRegistry, parsedDockerURL types.ParsedDockerURL, outputDir string) (path string, err error) {
 	util.Debug("Squashing layers...")
 	util.Debug("Rendering ACI...")
 	renderedACI, err := acirenderer.GetRenderedACIFromList(images, aciRegistry)
@@ -167,22 +166,34 @@ func SquashLayers(images []acirenderer.Image, aciRegistry acirenderer.ACIRegistr
 	}
 
 	squashedFilename := getSquashedFilename(parsedDockerURL)
-	squashedImagePath := path.Join(outputDir, squashedFilename)
+	squashedImagePath := filepath.Join(outputDir, squashedFilename)
 
-	squashedImageFile, err := os.Create(squashedImagePath)
+	squashedTempFile, err := ioutil.TempFile(outputDir, "docker2aci-squashedFile-")
 	if err != nil {
 		return "", err
 	}
-	defer squashedImageFile.Close()
+	defer func() {
+		if err == nil {
+			err = squashedTempFile.Close()
+		} else {
+			// remove temp file on error
+			// we ignore its error to not mask the real error
+			os.Remove(squashedTempFile.Name())
+		}
+	}()
 
 	util.Debug("Writing squashed ACI...")
-	if err := writeSquashedImage(squashedImageFile, renderedACI, aciRegistry, manifests); err != nil {
+	if err := writeSquashedImage(squashedTempFile, renderedACI, aciRegistry, manifests); err != nil {
 		return "", fmt.Errorf("error writing squashed image: %v", err)
 	}
 
 	util.Debug("Validating squashed ACI...")
-	if err := common.ValidateACI(squashedImagePath); err != nil {
+	if err := common.ValidateACI(squashedTempFile.Name()); err != nil {
 		return "", fmt.Errorf("error validating image: %v", err)
+	}
+
+	if err := os.Rename(squashedTempFile.Name(), squashedImagePath); err != nil {
+		return "", err
 	}
 
 	util.Debug("ACI squashed!")


lib/docker2aci.go
```diff
@@ -169,22 +169,27 @@ func SquashLayers(images []acirenderer.Image, aciRegistry acirenderer.ACIRegistr
...
 	squashedImagePath := path.Join(outputDir, squashedFilename)
 
-	squashedImageFile, err := os.Create(squashedImagePath)
+	squashedTempFile, err := ioutil.TempFile(outputDir, "squashedFilename-")
 	if err != nil {
 		return "", err
 	}
-	defer squashedImageFile.Close()
+	defer squashedTempFile.Close()
+	squashedTempFilePath := filepath.Join(outputDir, squashedTempFile.Name())
```
sgotti: Is `filepath.Join` really needed?
squashedTempFile.Name() should just contain the outputDir.

I think that with this, if outputDir is `/tmp/something/` squashedTempFile will become `/tmp/something/tmp/something/filename`

iaguis: You're right. I misinterpreted the `Name()` function.

lib/docker2aci.go
```diff
@@ -167,24 +166,28 @@ func SquashLayers(images []acirenderer.Image, aciRegistry acirenderer.ACIRegistr
 	}
 
 	squashedFilename := getSquashedFilename(parsedDockerURL)
-	squashedImagePath := path.Join(outputDir, squashedFilename)
+	squashedImagePath := filepath.Join(outputDir, squashedFilename)
 
-	squashedImageFile, err := os.Create(squashedImagePath)
+	squashedTempFile, err := ioutil.TempFile(outputDir, "squashedFilename-")
```
sgotti: Do you think that the tempFile should be removed on error?
Perhaps it can be created inside the tmpDir so it can be removed from who created it.

iaguis: tmpDir can be in another filesystem and the Rename call would fail, I guess we should defer a function that checks if there was any error and removes the temp file :/

sgotti: You're right!

iaguis: Updated with tmp file removal

lib/docker2aci.go
```diff
@@ -167,24 +166,39 @@ func SquashLayers(images []acirenderer.Image, aciRegistry acirenderer.ACIRegistr
...
 		return "", err
 	}
-	defer squashedImageFile.Close()
+	defer func() {
+		errClose := squashedTempFile.Close()
+		if err != nil {
+			// remove temp file on error
+			errRemove := os.Remove(squashedTempFile.Name())
+			if errRemove != nil {
+				err = errRemove
```
sgotti: Will this mask the "real" error returned to the caller and/or return an error also if writeSquashedImage/ValidateACI complete successfully but Close() failed?

iaguis: It will return the Remove error if there is one masking the "real" error.

If writeSquashedImage/ValidateACI complete successfully but Close() failed it will return the Close error.

Maybe we should return the "real" error instead of the Remove error? Not sure what's better.

sgotti: > Maybe we should return the "real" error instead of the Remove error? Not sure what's better.

I've got the same doubts. I'll ignore the Close and the Remove errors to not mask the "real" one. (for example, Close() errors are often ignored: `defer file.Close()`)

iaguis: I think it's important to not ignore the Close error on successful cases to avoid data corruption.

How about we remove the temp file on error ignoring the Remove errors so we don't mask the "real" one and we handle the close error if there was a successful execution?

```
defer func() {
	if err == nil {
		err = squashedTempFile.Close()
	} else {
		// remove temp file on error
		// we ignore its error to not mask the real error
		os.Remove(squashedTempFile.Name())
	}
}()
```

sgotti: > I think it's important to not ignore the Close error on successful cases to avoid data corruption.

You're right.


> How about we remove the temp file on error ignoring the Remove errors so we don't mask the "real" one and we handle the close error if there was a successful execution?

Seems Good!

iaguis:
Updated. PTAL

sgotti:
LGTM!

Issue #15, opened 2015-02-09
use docker image ID as value for version label


jonboulle:
Per OOB discussion, it would be helpful for people to correlate between Docker images and ACIs created through docker2aci by using the docker image's Image ID as the _version_ label in the produced ACI.

philips:
I am thinking we should standardize some of these details from the conversion process. Here are my suggestions:

- `appc.io/docker/v1/checksum`
- `appc.io/docker/v1/tag`
- `appc.io/docker/v1/repository`
- `appc.io/docker/v1/imageID`
- `appc.io/docker/v1/parentID`

iaguis:
Something like this makes total sense to me

philips:
Seems like we should have a WELLKNOWN.md file with well-known labels or something.

jonboulle:
We should loop in the quay.io guys here too, hmm?

josephschorr:
We currently add:

```
{"name": "quay.io/derived-image", "value": synthetic_image_id},
```

where `synthetic_image_id` is the image ID we generate for the squashed image. This allows us to map any ACI we generate back to an image, and thereafter back to a repository and tag.

iaguis:
Fixed by #78

PR #78, opened 2015-08-07
common: add additional docker informations to manifest.

sgotti:
This patch tries to implement the idea from #15.

This adds additional docker informations to the image manifest.
Now the v1 registry informations are added under the `appc.io/docker/v1/`
namespace:

`appc.io/docker/v1/indexurl`
`appc.io/docker/v1/repository`
`appc.io/docker/v1/tag`
`appc.io/docker/v1/imageid`
`appc.io/docker/v1/parentimageid`

Before adding a WELLKNOWN.md like proposed I'd like to get suggestions on the annotation names. For example `indexurl` should be `registryurl`?

This can be useful for getting docker v1 caching infos like proposed in https://github.com/coreos/rkt/issues/1196#issuecomment-127983674 without the need to save the imageID info in a store sql table.

### DIFF
--- a/lib/common/common.go
+++ b/lib/common/common.go
@@ -22,8 +22,13 @@ import (
 )
 
 const (
-	defaultTag    = "latest"
-	schemaVersion = "0.5.1"
+	defaultTag                = "latest"
+	schemaVersion             = "0.6.1"
+	appcDockerV1RegistryURL   = "appc.io/docker/v1/registryurl"
+	appcDockerV1Repository    = "appc.io/docker/v1/repository"
+	appcDockerV1Tag           = "appc.io/docker/v1/tag"
+	appcDockerV1ImageID       = "appc.io/docker/v1/imageid"
+	appcDockerV1ParentImageID = "appc.io/docker/v1/parentimageid"
 )
 
 func ParseDockerURL(arg string) *types.ParsedDockerURL {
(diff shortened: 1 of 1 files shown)


lib/common/common.go
```diff
@@ -22,8 +22,13 @@ import (
...
 const (
-	defaultTag    = "latest"
-	schemaVersion = "0.5.1"
+	defaultTag                = "latest"
+	schemaVersion             = "0.5.1"
+	AppcDockerV1IndexURL      = "appc.io/docker/v1/indexurl"
+	AppcDockerV1Repository    = "appc.io/docker/v1/repository"
+	AppcDockerV1Tag           = "appc.io/docker/v1/tag"
+	AppcDockerV1ImageID       = "appc.io/docker/v1/imageid"
+	AppcDockerV1ParentImageID = "appc.io/docker/v1/parentimageid"
```
iaguis: Why are they exported?

lib/common/common.go
```diff
@@ -22,8 +22,13 @@ import (
 )
 
 const (
-	defaultTag    = "latest"
-	schemaVersion = "0.5.1"
+	defaultTag                = "latest"
+	schemaVersion             = "0.5.1"
```
iaguis: While you're at it you can update schemaVersion :sweat_smile:

lib/common/common.go
sgotti: I was thinking to export them for use by rkt but this doesn't make much sense. I'll make them unexported.

lib/common/common.go
sgotti: Is 0.6.1 ok?

iaguis:
Just a small nit, can you `s/informations/information`, please?

Then it's good to merge.

sgotti:
@iaguis sure thanks! What do you think about the names (for example `indexurl` vs `registerurl`)?

iaguis:
Hmm... now that you mention it I think we should use `registryurl` since that's what we use in the documentation and what docker uses everywhere currently.

sgotti:
Done!

iaguis:
Cool. Thanks!

PR #80, opened 2015-08-12
common: update to new appc/spec aci.TarReadCloser.

sgotti:
Now we have to pass to functions wanting a *tar.Reader the embedded *tar.Reader
in aci.TarReadCloser and call the Close() method when finished.

See appc/spec#461


jonboulle:
I guess this is exactly the question I'd raised. Thanks!  Lgtm

iaguis:
Thanks! :+1:

PR #82, opened 2015-08-28
common: omit parent index URL if default index

iaguis:
We're doing that in generated image names so for dependencies to work we
need to do it too.


jonboulle:
LGTM

PR #84, opened 2015-09-16
Updated to use the new ioprogress library.

dgonyeo:
Switched bar drawing function so that the bar will be surpressed when
not drawing to a tty.


jonboulle:
surpressed -> suppressed

LGTM

dgonyeo:
surppressed (n): when something is suppressed in such a way as to be surprising.

fixed.

iaguis:
> surppressed (n): when something is suppressed in such a way as to be surprising.

:joy:

Issue #87, opened 2015-09-17
Fails to build with latest version of appc/spec


dgonyeo:
```
# github.com/appc/docker2aci/lib/common
lib/common/common.go:95: tr.Close undefined (type *tar.Reader has no field or method Close)
lib/common/common.go:97: tr.Reader undefined (type *tar.Reader has no field or method Reader)
lib/common/common.go:358: tr.Close undefined (type *tar.Reader has no field or method Close)
lib/common/common.go:360: tr.Reader undefined (type *tar.Reader has no field or method Reader)
```

Where tr comes from the `NewCompressedTarReader` function in `github.com/appc/spec/aci`

PR #88, opened 2015-09-17
The archive/tar package upstream changed

dgonyeo:
These fixes make it so that docker2aci can compile with the new version.

Fixes https://github.com/appc/docker2aci/issues/87

### DIFF
--- a/lib/common/common.go
+++ b/lib/common/common.go
@@ -92,9 +92,8 @@ func ValidateACI(aciPath string) error {
 	if err != nil {
 		return err
 	}
-	defer tr.Close()
 
-	if err := aci.ValidateArchive(tr.Reader); err != nil {
+	if err := aci.ValidateArchive(tr); err != nil {
 		return err
 	}
 
@@ -360,9 +359,8 @@ func writeACI(layer io.ReadSeeker, manifest schema.ImageManifest, curPwl []strin
 	}
 	tr, err := aci.NewCompressedTarReader(layer)
 	if err == nil {
-		defer tr.Close()
 		// write files in rootfs/
-		if err := tarball.Walk(*tr.Reader, convWalker); err != nil {
+		if err := tarball.Walk(*tr, convWalker); err != nil {
 			return nil, err
 		}
 	} else {


dgonyeo:
Builds under go 1.4 and 1.5

iaguis:
This is basically a revert of https://github.com/appc/docker2aci/commit/c675e7bd0ca7faf6b2f63014616c9358504b961b

[Once](https://github.com/coreos/rkt/pull/929#issuecomment-104787060) [again](https://github.com/appc/docker2aci/pull/65) we're in the unfortunate situation where the master branch of appc/spec changed something and rkt uses an old version.

If we don't apply your PR, rkt can't build with the last version of docker2aci vendored; if we do, the docker2aci build is broken.

We can apply this PR and then revert it so rkt can vendor a working version of docker2aci and we don't break the build (this is what we did in #65).

Maybe having a long-lived `rkt` branch in docker2aci which we keep synced to the appc version rkt is vendoring would be a better solution...

jonboulle:
> If we don't apply your PR, rkt can't build with the last version of docker2aci vendored; if we do, the docker2aci build is broken.

@iaguis this is inverted, right? I'm confused

iaguis:
Right now docker2aci can be built against the latest spec but rkt has v0.6.1 vendored. If we want to use the master branch of docker2aci in rkt we need to merge this PR which makes docker2aci compatible with appc/spec v0.6.1. But then we'd break building docker2aci against the latest spec.

Does it make sense now? Maybe I'm missing something.

jonboulle:
OK, sorry for the confusion, I was thinking backwards. For my own edification: this is not because of an archive/tar change but rather the `appc/spec/aci` changes in https://github.com/appc/spec/pull/461/files

jonboulle:
I think the solution here is for everyone to stop at nothing until I do an appc release.

dgonyeo:
I believe I was trying to build it with an old version of appc/spec (thanks, godep), and thus noticed a nonexistent problem. It appears to build just fine with master from appc/spec.

Nevermind.

PR #95, opened 2015-10-02
lib: fix multiple rootfs entries

### LINKED ISSUE #94: Generated ACI contains several entries for "rootfs"
In this example, 3 entries for "rootfs":
```
$ docker2aci docker://busybox
$ tar tvzf busybox-latest.aci|grep rootfs$
drwxr-xr-x 0/0               0 1970-01-01 01:00 rootfs
drwxr-xr-x 0/0               0 1970-01-01 01:00 rootfs
drwxr-xr-x 0/0               0 1970-01-01 01:00 rootfs
$
```

I discovered this when I wanted to check the permissions of `/`. It does not seem to be causing any visible problems.

iaguis:
acirenderer adds a rootfs entry since
https://github.com/appc/spec/pull/437

Hence, we need to ignore it because we write our own rootfs entry.

Fixes #94

### DIFF
--- a/lib/docker2aci.go
+++ b/lib/docker2aci.go
@@ -241,8 +241,8 @@ func writeSquashedImage(outputFile *os.File, renderedACI acirenderer.RenderedACI
 			cleanName := filepath.Clean(t.Name())
 
 			if _, ok := aciFile.FileMap[cleanName]; ok {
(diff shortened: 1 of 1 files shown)


jonboulle:
LGTM

Issue #46, opened 2015-04-16
Make docker2aci compatible with the new Docker Registry 2.0.0


alban:
The [new Docker Registry 2.0.0](https://groups.google.com/a/dockerproject.org/forum/#!msg/distribution/ua7dpanYmhU/g_zIv4SR4ioJ) is getting ready.

We should check whether docker2aci works fine with it and investigate what changes in docker2aci are necessary to support it.

dgonyeo:
I've spent a lot of time trying to work on this, but I've ended up getting stuck on auth in the new API. I'm going to write down my current state and issue, and hopefully someone else can tell me where things are going wrong.

Any requests I make against the v2 registry at `registry-1.docker.io` return a 401 Unauthorized, which is expected. It sets the `www-authenticate` header to something like:

```
Www-Authenticate: Bearer realm="https://auth.docker.io/token",service="registry.docker.io",scope="repository:centos:pull"
```

According to [the docs](https://docs.docker.com/registry/spec/auth/token/) that describe the auth process, after getting this 401 error the client should go get an auth token based on that header. This just requires making a GET request against the realm specified, with the get parameters service and scope set to the values from the header.

```
curl -v "https://auth.docker.io/token?service=registry.docker.io&scope=repository:coreos:pull"
```

That results in a token being returned (what the token contains is documented [here](https://docs.docker.com/registry/spec/auth/jwt/)). There are 3 sections, separated by a `.`, and the second one is a base64 encoded json blob, in which it'll describe what things the token is good for accessing.

Every token I've gotten has had the access array in that 2nd section empty, which means it doesn't have permission to access anything, and using it just results in more 401 Unauthorized responses with identical `www-authenticate` headers set.

stevenschlansker:
Two things to consider -- one big use case is unauthenticated private registries, for which the authentication is not relevant.  So maybe that part can be finished first, independent of public registry support.
Second, maybe this is a good question to post to either the docker/distribution repo or the mailing lists?

dgonyeo:
Are you aware of an unauthenticated registry I could test this against? I'm pretty sure I could get this working without too much trouble if the auth was irrelevant.

I guess I'll send something to one of the mailing lists, hopefully someone will know what's up.

stevenschlansker:
I'm not aware of any publicly accessible ones, but running a test instance locally is [nearly trivial](https://github.com/docker/distribution/blob/master/docs/deploying.md)

jonboulle:
[...]

Issue #89, opened 2015-09-17
Docker images with PaxHeaders generate ACI images with unequal hashes


blixtra:
When fetching the same Docker image one should get an ACI image with identical hashes. With docker://busybox for example this is the case.

```
~/W/s/rkt: sudo rkt --insecure-skip-verify fetch docker://busybox
Downloading cf2616975b4a: [====================================] 32 B/32 B
Downloading 6ce2e90b0bc7: [====================================] 1.15 MB/1.15 MB
Downloading 8c2e06607696: [====================================] 32 B/32 B
sha512-9d710100ce6769569b12a39100318bfe
~/W/s/rkt: sudo rkt --insecure-skip-verify fetch docker://busybox
Downloading cf2616975b4a: [====================================] 32 B/32 B
Downloading 6ce2e90b0bc7: [====================================] 1.15 MB/1.15 MB
Downloading 8c2e06607696: [====================================] 32 B/32 B
sha512-9d710100ce6769569b12a39100318bfe
```

However, with the redis image this is sadly not the case.

```
~/W/s/rkt: sudo rkt --insecure-skip-verify fetch docker://redis
Downloading ba249489d0b6: [====================================] 37.2 MB/37.2 MB
Downloading 19de96c112fc: [====================================] 32 B/32 B
Downloading d990a769a35e: [====================================] 1.7 KB/1.7 KB
Downloading 8656a511ce9c: [====================================] 5.94 MB/5.94 MB
Downloading f7022ac152fb: [====================================] 93.6 KB/93.6 KB
Downloading 8e84d9ce7554: [====================================] 611 KB/611 KB 
Downloading c9e5dd2a9302: [====================================] 32 B/32 B
Downloading 27b967cdd519: [====================================] 32 B/32 B
Downloading 3024bf5093a1: [====================================] 32 B/32 B
Downloading e6a9eb403efb: [====================================] 3.04 MB/3.04 MB
Downloading c3532a4c89bc: [====================================] 98 B/98 B
Downloading 35fc08946add: [====================================] 32 B/32 B
Downloading d586de7d17cd: [====================================] 32 B/32 B
Downloading 1f677d77a8fa: [====================================] 196 B/196 B
Downloading ed09b32b8ab1: [====================================] 32 B/32 B
Downloading 54647d88bc19: [====================================] 32 B/32 B
Downloading 2f2578ff984f: [====================================] 32 B/32 B
sha512-039afdb1311dc35a802a48352b6b29f2
[...]

PR #99, opened 2015-11-11
adds support for the V2 HTTP API

dgonyeo:
Will detect when fetching an image if the V2 API is supported, and fall
back to the V1 API if it is not. Auth currently unsupported.

Partially addresses https://github.com/appc/docker2aci/issues/46 and https://github.com/coreos/rkt/issues/1583.

### DIFF
--- a/lib/backend/repository/repository1.go
+++ b/lib/backend/repository/repository1.go
@@ -1,0 +1,350 @@
+// Copyright 2015 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package repository
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"path"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/appc/docker2aci/lib/common"
+	"github.com/appc/docker2aci/lib/types"
+	"github.com/appc/docker2aci/lib/util"
+	"github.com/appc/spec/schema"
+	"github.com/coreos/ioprogress"
+)
+
+type RepoData struct {
+	Tokens    []string
+	Endpoints []string
+	Cookie    []string
+}
+
+func (rb *RepositoryBackend) getImageInfoV1(dockerURL *types.ParsedDockerURL) ([]string, *types.ParsedDockerURL, error) {
+	repoData, err := rb.getRepoDataV1(dockerURL.IndexURL, dockerURL.ImageName)
+	if err != nil {
+		return nil, nil, fmt.Errorf("error getting repository data: %v", err)
+	}
+
+	// TODO(iaguis) check more endpoints
+	appImageID, err := rb.getImageIDFromTagV1(repoData.Endpoints[0], dockerURL.ImageName, dockerURL.Tag, repoData)
+	if err != nil {
+		return nil, nil, fmt.Errorf("error getting ImageID from tag %s: %v", dockerURL.Tag, err)
+	}
+
+	ancestry, err := rb.getAncestryV1(appImageID, repoData.Endpoints[0], repoData)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	rb.repoData = repoData
+
+	return ancestry, dockerURL, nil
+}
+
+func (rb *RepositoryBackend) buildACIV1(layerNumber int, layerID string, dockerURL *types.ParsedDockerURL, outputDir string, tmpBaseDir string, curPwl []string, compress bool) (string, *schema.ImageManifest, error) {
+	tmpDir, err := ioutil.TempDir(tmpBaseDir, "docker2aci-")
+	if err != nil {
+		return "", nil, fmt.Errorf("error creating dir: %v", err)
+	}
+	defer os.RemoveAll(tmpDir)
+
+	j, size, err := rb.getJsonV1(layerID, rb.repoData.Endpoints[0], rb.repoData)
+	if err != nil {
+		return "", nil, fmt.Errorf("error getting image json: %v", err)
+	}
+
+	layerData := types.DockerImageData{}
+	if err := json.Unmarshal(j, &layerData); err != nil {
+		return "", nil, fmt.Errorf("error unmarshaling layer data: %v", err)
+	}
+
+	layerFile, err := rb.getLayerV1(layerID, rb.repoData.Endpoints[0], rb.repoData, size, tmpDir)
+	if err != nil {
+		return "", nil, fmt.Errorf("error getting the remote layer: %v", err)
+	}
+	defer layerFile.Close()
+
+	util.Debug("Generating layer ACI...")
+	aciPath, manifest, err := common.GenerateACI(layerNumber, layerData, dockerURL, outputDir, layerFile, curPwl, compress)
+	if err != nil {
+		return "", nil, fmt.Errorf("error generating ACI: %v", err)
+	}
+
+	return aciPath, manifest, nil
+}
+
+func (rb *RepositoryBackend) getRepoDataV1(indexURL string, remote string) (*RepoData, error) {
+	client := &http.Client{}
+	repositoryURL := rb.protocol() + path.Join(indexURL, "v1", "repositories", remote, "images")
+
+	req, err := http.NewRequest("GET", repositoryURL, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	if rb.username != "" && rb.password != "" {
+		req.SetBasicAuth(rb.username, rb.password)
+	}
+
+	req.Header.Set("X-Docker-Token", "true")
+
+	res, err := client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	defer res.Body.Close()
+
+	if res.StatusCode != 200 {
+		return nil, fmt.Errorf("HTTP code: %d, URL: %s", res.StatusCode, req.URL)
+	}
+
+	var tokens []string
+	if res.Header.Get("X-Docker-Token") != "" {
+		tokens = res.Header["X-Docker-Token"]
+	}
+
+	var cookies []string
+	if res.Header.Get("Set-Cookie") != "" {
+		cookies = res.Header["Set-Cookie"]
+	}
+
+	var endpoints []string
+	if res.Header.Get("X-Docker-Endpoints") != "" {
+		endpoints = makeEndpointsListV1(res.Header["X-Docker-Endpoints"])
+	} else {
+		// Assume same endpoint
+		endpoints = append(endpoints, indexURL)
+	}
+
+	return &RepoData{
+		Endpoints: endpoints,
+		Tokens:    tokens,
+		Cookie:    cookies,
+	}, nil
+}
+
+func (rb *RepositoryBackend) getImageIDFromTagV1(registry string, appName string, tag string, repoData *RepoData) (string, error) {
+	client := &http.Client{}
+	// we get all the tags instead of directly getting the imageID of the
+	// requested one (.../tags/TAG) because even though it's specified in the
+	// Docker API, some registries (e.g. Google Container Registry) don't
+	// implement it.
+	req, err := http.NewRequest("GET", rb.protocol()+path.Join(registry, "repositories", appName, "tags"), nil)
+	if err != nil {
+		return "", fmt.Errorf("failed to get Image ID: %s, URL: %s", err, req.URL)
+	}
+
+	setAuthTokenV1(req, repoData.Tokens)
+	setCookieV1(req, repoData.Cookie)
+	res, err := client.Do(req)
+	if err != nil {
+		return "", fmt.Errorf("failed to get Image ID: %s, URL: %s", err, req.URL)
+	}
+	defer res.Body.Close()
+
+	if res.StatusCode != 200 {
+		return "", fmt.Errorf("HTTP code: %d. URL: %s", res.StatusCode, req.URL)
+	}
+
+	j, err := ioutil.ReadAll(res.Body)
+
+	if err != nil {
+		return "", err
+	}
+
+	var tags map[string]string
+
+	if err := json.Unmarshal(j, &tags); err != nil {
+		return "", fmt.Errorf("error unmarshaling: %v", err)
+	}
+
+	imageID, ok := tags[tag]
+	if !ok {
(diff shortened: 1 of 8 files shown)


dgonyeo:
As of right now all the information for the given layers comes from the "v1Compatibility" field, which I feel really uneasy about. The schema for the manifest doesn't say if it's an optional field, if it needs to conform to the schema of the v1 API, or much else about it.

The [example manifest](https://github.com/docker/distribution/blob/master/docs/spec/manifest-v2-1.md#example-manifest) in their documentation doesn't even have the correct number of elements in the history array to be able to correlate each one to a layer in the image. I can't find anywhere else in the API that contains the information in it though (the information is what would have been returned in the V1 API by a `GET /v1/images/(image_id)/json` request [here](https://docs.docker.com/v1.6/reference/api/registry_api/).

So either this field is poorly documented and their example manifest is malformed, or there's some way to get the necessary information about each layer in the V2 API I'm missing.

jonboulle:
Can you ping upstream about this to check? #docker-distribution on IRC should be the place to start.

dgonyeo:
Will do

lib/backend/repository/repository2.go
```diff
@@ -0,0 +1,254 @@
...
+	FSLayers []struct {
+		BlobSum string `json:"blobSum"`
+	} `json:"fsLayers"`
+	History []struct {
+		V1Compatibility string `json:"v1Compatibility"`
+	} `json:"history"`
+	//Signature []byte `json:"signature"`
+}
+
+func (rb *RepositoryBackend) supportsV2(indexURL string) (bool, error) {
```
iaguis: The [API spec](https://docs.docker.com/registry/spec/api/) mentions this:

> When a 200 OK or 401 Unauthorized response is returned, the “Docker-Distribution-API-Version” header should be set to “registry/2.0”. Clients may require this header value to determine if the endpoint serves this API. When this header is omitted, clients may fallback to an older API version.

I understand that we're not supporting auth for now but maybe we should check that header although a 200 OK is probably a good enough indication that v2 is supported.

iaguis:
Missing license headers.

iaguis:
Since auth is not supported, is there any v2 registry to test this against?

dgonyeo:
There isn't a public one, but it's easy enough to run your own:
https://github.com/docker/distribution/blob/master/docs/deploying.md

lib/common/common.go
```diff
@@ -107,9 +107,9 @@ func GenerateManifest(layerData types.DockerImageData, dockerURL *types.ParsedDo
 
 	appURL := ""
 	// omit docker hub index URL in app name
-	if dockerURL.IndexURL != defaultIndex {
-		appURL = dockerURL.IndexURL + "/"
-	}
+	//if dockerURL.IndexURL != defaultIndex {
```
iaguis: If we need to remove the defaultIndex because we have several kinds now, we can remove this instead of commenting it out.

iaguis:
> There isn't a public one, but it's easy enough to run your own:
> https://github.com/docker/distribution/blob/master/docs/deploying.md

Thanks!

The PR looks good in general.

Some comments:

* We still use the v1 annotations (`appc.io/docker/v1/registryurl` and so on) even if we get the image from a v2 registry.
* The `--insecure` flag is described as "Allows unencrypted connections when fetching images". As far as I can see it forces unencrypted connections by using `http://`. Also, it should be added to the example string (`docker2aci [--debug] [--nosquash] IMAGE`).
* When trying to connect to a v2 that doesn't support TLS, the error message is not very descriptive: `Error: conversion error: Get https://localhost:5000/v2: tls: oversized record received with length 20527`, perhaps we should detect that the server doesn't support it and return a proper error.
* We can't merge this before we suppor auth because it breaks pulling images from the Docker Hub:

 ```
 > docker2aci docker://busybox
 Error: conversion error: unexpected http code: 401, URL: https://registry-1.docker.io/v2/busybox/manifests/latest
 ```

jonboulle:
any luck badgering upstream about auth?

dgonyeo:
Still on my to do list :p

dgonyeo:
I asked in IRC about the v1Compatibility thing I described earlier in this thread, and docker is "working on an improved manifest format". Until that's done, we "need to use v1Compatibility". So expect docker2aci needing to be modified in the future for their new format, but that's the field we want to get this information from for now.

dgonyeo:
I figured out the auth issue. Turns out my curl commands failed because any repo named like `centos` is actually `library/centos`. I just updated the commit to have the auth stuff, so please review the changes.

There is an issue atm though, where some images will spit out the following error:

```
Error: conversion error: error squashing image: error rendering squashed image: image hash does not match expected (sha512-5b1aa5a55395cb7bd9f9c0a0d247908a82c406e8da7690a04be2d5ed71d16784f33c0588ddb3396c168d816d6b3da2f4c5e45dd81d2b4d8ba001fd026e17eab7 != sha512-bbcacd83a85b6b0157de97ceb691d797ef4f0bebbfef1f31edb5c5fa57376880c866f3955453a8523e7b1507a68b13a2cfe72190f9eec1bd0ae1446690c49906)
```

I'm looking in to that now. For anyone who wants to test auth against the public registry, I've been able to successfully convert the `alpine` and `ubuntu` images.

jonboulle:
@iaguis any ideas on that error?

iaguis:
Hm...

We import the converted layers to a in-memory conversion_store that hashes the ACI, generates a key from the hash and has functions to return the ACI file path by key.

That error means that the stored hash doesn't match the actual hash of the file. Which is very weird, we don't do anything fancy in [conversion_store.go](https://github.com/appc/docker2aci/blob/fce83d1375e33a840e455f51a102491111984924/lib/conversion_store.go).

iaguis:
Some layers have the same `sha256` so they end up with the same name and overwrite each other:

```
> docker2aci --nosquash docker://quay.io/coreos/etcd
Downloading sha256:f05e5: [====================================] 3.98 MB/3.98 MB
Downloading sha256:af189: [====================================] 3.5 MB/3.5 MB 
Downloading sha256:a3ed9: [====================================] 32 B/32 B
Downloading sha256:a3ed9: [====================================] 32 B/32 B

Converted ports:
	name: "2379-tcp", protocol: "tcp", port: 2379, count: 1, socketActivated: false
	name: "2380-tcp", protocol: "tcp", port: 2380, count: 1, socketActivated: false
	name: "4001-tcp", protocol: "tcp", port: 4001, count: 1, socketActivated: false
	name: "7001-tcp", protocol: "tcp", port: 7001, count: 1, socketActivated: false

Generated ACI(s):
coreos-etcd-d21dd9a5886270b7c2c379c02fc548e0696b139c43bb12fdb2d9b63409717485-latest-linux-amd64.aci
coreos-etcd-620329641f386e62c7b0e0fa60a9acef100e71058124ddc7f1969557c72b2458-latest-linux-amd64.aci
coreos-etcd-9cd3f08f7ccfaad24c73757a5b4f79601f2790726d6ccdd556a82e5c9c5ddbfa-latest-linux-amd64.aci
coreos-etcd-9cd3f08f7ccfaad24c73757a5b4f79601f2790726d6ccdd556a82e5c9c5ddbfa-latest-linux-amd64.aci
> ls
coreos-etcd-620329641f386e62c7b0e0fa60a9acef100e71058124ddc7f1969557c72b2458-latest-linux-amd64.aci
coreos-etcd-9cd3f08f7ccfaad24c73757a5b4f79601f2790726d6ccdd556a82e5c9c5ddbfa-latest-linux-amd64.aci
coreos-etcd-d21dd9a5886270b7c2c379c02fc548e0696b139c43bb12fdb2d9b63409717485-latest-linux-amd64.aci
```

iaguis:
Wild guessing here: could it be that because the images are now content-addressable and both are the same (empty), only one layer is served?

Trying with other images, it seems the empty layers have the same `sha256` (note the `32 B` images):

```
Downloading sha256:6ffe5: [====================================] 65.8 MB/65.8 MB
Downloading sha256:f4e00: [====================================] 71.5 KB/71.5 KB
Downloading sha256:e99f3: [====================================] 680 B/680 B
Downloading sha256:a3ed9: [====================================] 32 B/32 B
Downloading sha256:a3ed9: [====================================] 32 B/32 B
Downloading sha256:a3ed9: [====================================] 32 B/32 B
Downloading sha256:7e1e5: [====================================] 83.2 MB/83.2 MB
Downloading sha256:4c72f: [====================================] 167 B/167 B
Downloading sha256:881ef: [====================================] 135 B/135 B
Downloading sha256:43f0b: [====================================] 438 B/438 B
Downloading sha256:e9421: [====================================] 545 B/545 B
Downloading sha256:9ad12: [====================================] 203 B/203 B
Downloading sha256:731c8: [====================================] 242 B/242 B
Downloading sha256:af655: [====================================] 243 B/243 B
Downloading sha256:e85be: [====================================] 277 B/277 B
Downloading sha256:728f7: [====================================] 626 B/626 B
Downloading sha256:bb6b0: [====================================] 1.01 KB/1.01 KB
Downloading sha256:9c9e5: [====================================] 680 B/680 B
Downloading sha256:bd75f: [====================================] 295 B/295 B
Downloading sha256:47d9c: [====================================] 53.1 KB/53.1 KB
Downloading sha256:0be50: [====================================] 144 B/144 B
Downloading sha256:a3ed9: [====================================] 32 B/32 B
Downloading sha256:a3ed9: [====================================] 32 B/32 B
Downloading sha256:a3ed9: [====================================] 32 B/32 B
Downloading sha256:a3ed9: [====================================] 32 B/32 B
Downloading sha256:a3ed9: [====================================] 32 B/32 B
Downloading sha256:a3ed9: [====================================] 32 B/32 B
Error: conversion error: error squashing image: error rendering squashed image: image hash does not match expected (sha512-a4062020e8770a42626219c3cb10c3adc03e21b045fe9da06aee76a484ba99fcae979ea2740d5247e5a3968b13731c8913a764ff3b3153a986df3741447a37b0 != sha512-1d540e7e737ca872995b8cf610a811119ca473941b7caa136fe25f19e362e97534a9c43baa28175ca4119f85ed0f44f3c66ab4dddac4bf17b39fe7be85f880c5)
```

iaguis:
I added a "random" number to the end of the file name and I can't reproduce the error anymore.

iaguis:
This wouldn't be a problem if we could generate reproducible images. This is not the case now because 

* Golang maps are randomized: the order of the fields in the manifest can be different on each execution.
* https://github.com/appc/docker2aci/issues/89

jonboulle:
I'm not sure about that pax one, but at least on the manifest side I would
LOVE to see patches to serialize them deterministically..

iaguis:
> I'm not sure about that pax one, but at least on the manifest side I would
> LOVE to see patches to serialize them deterministically..

I have some somewhere, I'll clean them up :)

iaguis:
https://github.com/appc/docker2aci/pull/107

jonboulle:
Oh, I was thinking of the appc serialization code but is the answer there
that it's already deterministicish because the go json package happens to
serialize maps in order?

Does this mean this should now work but for the pathological pax cases?

Iago López Galeiras <notifications@github.com> schrieb am Di., 1. Dez. 2015
01:09:

dgonyeo:
I just applied @iaguis's patch. I'd personally prefer to instead just number the layers and append the layer's number (as in 0, 1, 2, ...), but knowing the number of what layer it's working on would be messy there.

Seems to work for me fully now, should be ready for another pass.

iaguis:
For the record, my patch was a proof of concept and I also prefer numbering the layers.

iaguis:
> Oh, I was thinking of the appc serialization code but is the answer there
> that it's already deterministicish because the go json package happens to
> serialize maps in order?

AFAICT, there're no maps in the spec. The problem was that Docker uses maps for ports and volumes which get randomized, causing the converted manifest to be non-deterministisch.

> Does this mean this should now work but for the pathological pax cases?

I think so, yes.

dgonyeo:
The layers are now numbered in order, instead of randomly.

lib/backend/repository/repository2.go
```diff
@@ -0,0 +1,370 @@
...
+type v2Manifest struct {
+	Name     string `json:"name"`
+	Tag      string `json:"tag"`
+	FSLayers []struct {
+		BlobSum string `json:"blobSum"`
+	} `json:"fsLayers"`
+	History []struct {
+		V1Compatibility string `json:"v1Compatibility"`
+	} `json:"history"`
+	//Signature []byte `json:"signature"`
```
iaguis: You can uncomment this even if you don't use it (yet).

lib/common/common.go
```diff
@@ -208,9 +217,9 @@ func GenerateManifest(layerData types.DockerImageData, dockerURL *types.ParsedDo
 	if layerData.Parent != "" {
 		indexPrefix := ""
 		// omit docker hub index URL in app name
-		if dockerURL.IndexURL != defaultIndex {
-			indexPrefix = dockerURL.IndexURL + "/"
-		}
+		//if dockerURL.IndexURL != defaultIndex {
+		indexPrefix = dockerURL.IndexURL + "/"
+		//}
```
iaguis: Just remove those instead of commenting them out (also the comment above).

lib/backend/repository/repository1.go
```diff
@@ -0,0 +1,351 @@
...
+
+func makeEndpointsListV1(headers []string) []string {
+	var endpoints []string
+
+	for _, ep := range headers {
+		endpointsList := strings.Split(ep, ",")
+		for _, endpointEl := range endpointsList {
+			endpoints = append(
+				endpoints,
+				// TODO(iaguis) discover if httpsOrHTTP
```
iaguis: I guess this comment doesn't make sense anymore since we're using the `--insecure` flag to make it `http` or `https`

main.go
```diff
@@ -32,9 +32,10 @@ var (
 	flagNoSquash = flag.Bool("nosquash", false, "Don't squash layers and output every layer as ACI")
 	flagImage    = flag.String("image", "", "When converting a local file, it selects a particular image to convert. Format: IMAGE_NAME[:TAG]")
 	flagDebug    = flag.Bool("debug", false, "Enables debug messages")
+	flagInsecure = flag.Bool("insecure", false, "Allows unencrypted connections when fetching images")
```
iaguis: More than allowing it's forcing unencrypted connections: if it's set it will use `http://`

lib/backend/repository/repository2.go
```diff
@@ -0,0 +1,370 @@
...
+	var in io.Reader
+	in = res.Body
+
+	if hdr := res.Header.Get("Content-Length"); hdr != "" {
+		imgSize, err := strconv.ParseInt(hdr, 10, 64)
+		if err != nil {
+			return nil, err
+		}
+
+		prefix := "Downloading " + layerID[:12]
```
iaguis: We can increase a bit the length of the printed layerID because it includes `sha256:` now.

iaguis:
Cool, it works! :D 

Some nits but after fixing them LGTM.

main.go
dgonyeo: I'll change the wording, but the server could still redirect to an https connection.

iaguis: Right.

dgonyeo:
Nits should be fixed. Want to give this a final :+1:?

iaguis:
LGTM :+1:

Issue #98, opened 2015-10-21
handle hard links and whiteouts in different layers


alban:
Bug initially reported on https://github.com/coreos/rkt/issues/1653: if a Docker image has the hard links `/LINK-A` and `/LINK-B` in the first layer but `/LINK-A` gets deleted in the second layer, `docker2aci` generates a squashed tarball that contains only `/LINK-B` as a dangling reference, so the tarball cannot be extracted.

To produce the image [docker://albanc/busybox-hardlinks](https://hub.docker.com/r/albanc/busybox-hardlinks/), I followed the steps:
```
# docker run -ti busybox
(container)# touch /LINK-A
(container)# ln /LINK-A /LINK-Z
(container)# exit
# docker commit $container_id
# docker run -ti $container_id
(container)# rm -f /LINK-A
(container)# exit
# docker commit $container_id_2 albanc/busybox-hardlinks
$ docker2aci docker://albanc/busybox-hardlinks
$ tar xf albanc-busybox-hardlinks-latest.aci
tar: rootfs/LINK-Z: Cannot hard link to ‘rootfs/LINK-A’: No such file or directory
```

alban:
/cc @josephschorr @jakedt

The same bug exists in the quay.io Docker image squasher. I uploaded my [busybox-hardlinks Docker image](https://hub.docker.com/r/albanc/busybox-hardlinks/) into [quay.io/alban/busybox-hardlinks](https://quay.io/repository/alban/busybox-hardlinks) and then downloaded a squashed image from [the quay.io web interface](https://quay.io/repository/alban/busybox-hardlinks?tab=tags) (click on Fetch tag / Squashed Docker image). Then, the image also has a dangling hard link, so it cannot be extracted:

```
$ tar tvf /tmp/quay.io-alban-busybox-hardlinks-latest.squash.aci | grep LINK
hrw-r--r-- 0/0               0 2015-10-21 23:26 rootfs/LINK-Z link to rootfs/LINK-A
$ tar xf /tmp/quay.io-alban-busybox-hardlinks-latest.squash.aci
tar: rootfs/LINK-Z: Cannot hard link to ‘rootfs/LINK-A’: No such file or directory
```

josephschorr:
@alban: I just tried this locally without any errors. I pulled the image from DockerHub, pushed to my local Quay, then squash pulled image. Do you have another repro case I can try?

alban:
[...]

PR #100, opened 2015-11-12
handle hard links and whiteouts correctly

alban:
Hard links and whiteout were not handled correctly in this scenario:
- The first Docker layer contains:
  - file A
  - file B hard link to file A
- The second Docker layer contains:
  - whiteout file A

Then, the squashed ACI previously generated by docker2aci contained:
  - file B dangling hard link to file A

Hence the bug.

This patch changes:
- the converting algorithm lib/common/common.go:writeACI() to generate:
  - First docker layer
    - .hidden.docker2aci.sha512-xxx (hash of the layer and file name)
    - file A hard link to .hidden.docker2aci.sha512-xxx
    - file B hard link to .hidden.docker2aci.sha512-xxx
  - Second docker layer
    - whiteout file A
- the squashing algorithm lib/docker2aci.go:SquashLayers() to have two
  passes:
  - Pass one: build an in-memory map of hard links and whiteouts
  - Pass two: remove white-out and .hidden.docker2aci.sha512-xxx files

I tested the following images from https://github.com/coreos/rkt/issues/1653:
- docker://albanc/busybox-hardlinks
- docker://zopyx/xmldirector-plone

TODO:
- not optimized if there are no hard links
- semantic changes on --nosquash

Fixes https://github.com/appc/docker2aci/issues/98

### DIFF
--- a/lib/docker2aci.go
+++ b/lib/docker2aci.go
@@ -230,6 +230,32 @@ func writeSquashedImage(outputFile *os.File, renderedACI acirenderer.RenderedACI
 	outputWriter := tar.NewWriter(gw)
 	defer outputWriter.Close()
 
+	finalManifest := mergeManifests(manifests)
+
+	if err := common.WriteManifest(outputWriter, finalManifest); err != nil {
+		return err
+	}
+
+	if err := common.WriteRootfsDir(outputWriter); err != nil {
+		return err
+	}
+
+	type hardLinkEntry struct {
+		firstLinkCleanName string
+		firstLinkHeader    tar.Header
+		walked             bool
+	}
+	hardLinks := make(map[string]hardLinkEntry)
+
+	type tempEntry struct {
+		cleanName string
+		keep      bool
+		firstLink bool
+	}
+	var entries []tempEntry
+
+	// first pass: read all the entries and build tempEntry, hardLinks in memory
+	// but don't write on disk
 	for _, aciFile := range renderedACI {
 		rs, err := aciProvider.ReadStream(aciFile.Key)
 		if err != nil {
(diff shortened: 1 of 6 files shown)


alban:
@iaguis I think this is ready for a first review.

lib/docker2aci.go
```diff
@@ -224,12 +224,42 @@ func getManifests(renderedACI acirenderer.RenderedACI, aciRegistry acirenderer.A
 	return manifests, nil
 }
 
-func writeSquashedImage(outputFile *os.File, renderedACI acirenderer.RenderedACI, aciProvider acirenderer.ACIProvider, manifests []schema.ImageManifest) error {
+func writeSquashedImage(outputFile *os.File,
+	renderedACI acirenderer.RenderedACI,
+	aciProvider acirenderer.ACIProvider,
+	manifests []schema.ImageManifest) error {
```
iaguis: You don't touch the propotype anymore so you can leave the line as it was

iaguis:
* Right now it only handles hardlinks to regular files, although weird, you could have a hardlink to a symlink
* I'm not concerned about this not being optimal for the no-hardlink case, but I am about the semantic changes in `--nosquash`. You mention that as a TODO, do you have any plans for it?
 
 Actually, it doesn't work. I just tested converting a busybox image with `--nosquash` and trying to run it in rkt results in

 ```
run: error setting up stage0: error setting up image \
sha512-2dffd725e98855c4afd53bbe2f053f7f3e94565fda1c7f6b753729998a997f47: \
error rendering tree image: treestore: cannot render aci: error extracting ACI: \
extracttar error: exit status 1, output: error extracting tar: error extracting tarball: link \
/rootfs/.hidden.docker2aci.sha512-fa4bf224598acc71239bc5ae543e9fa38f6f9032d2d219be2ba79f7a54259f505dcddc1b2ce3d8b643c8e83a356b24e837ced7637dc093d92bc88792af9c1947 \
/rootfs/bin/[: no such file or directory
 ```
 That happens because `.hidden.docker2aci.sha512-...` is not in the path white list so it doesn't get extracted and the hardlinks pointing to it are dangling.
 I don't think we should merge it in the current state because of that.

alban:
Closing in favor of #105

PR #101, opened 2015-11-12
lib: don't close reader inside ReadStream

iaguis:
When aci.NewCompressedReader was changed to return a ReadCloser, we
changed the code in ReadStream to close it and then wrapped it with a
NopCloser. That was obviously wrong although it didn't seem to have any
effect.

Let's return the ReadCloser directly instead.


jonboulle:
yikes. Great catch.

PR #103, opened 2015-11-20
tests: add tests for CI in Semaphore

alban:


iaguis:
LGTM

tests/test-basic/check.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+set -e
+grep -q file1 file1
+grep -q file1 file2
+grep -q file3 file3
+grep -q file4 file4
+if [ "$rendered" != "true" ] ; then
```
alban: For the record, this is to skip the test for the `rkt image render` command because of https://github.com/coreos/rkt/issues/1774

PR #105, opened 2015-11-23
handle hard links and whiteouts correctly (v3)

alban:
Replace #100 to trigger the tests in Semaphore.
Compared to the previous version:
- no changes with `--nosquash`: no `.hidden` files anymore
- include tests

-----
Hard links and whiteout were not handled correctly in this scenario:
- The first Docker layer contains:
  - file A
  - file B hard link to file A
- The second Docker layer contains:
  - whiteout file A

Then, the squashed ACI previously generated by docker2aci contained:
  - file B dangling hard link to file A

Hence the bug.

This patch changes the squashing algorithm lib/docker2aci.go:SquashLayers()
to have two passes:
  - Pass one: build an in-memory map of hard links and whiteouts
  - Pass two: write the archive using the map generated in pass one

I tested the following images from https://github.com/coreos/rkt/issues/1653:
- docker://albanc/busybox-hardlinks
- docker://zopyx/xmldirector-plone

Fixes https://github.com/appc/docker2aci/issues/98

### DIFF
--- a/tests/test-whiteouts/check.sh
+++ b/tests/test-whiteouts/check.sh
@@ -1,0 +1,66 @@
+#!/bin/sh
+set -e
+set -x
+
+grep -q yes layer0-file1
+grep -q yes layer0-file2
+grep -q yes layer0-file3
+
+test ! -e layer1-file1
+test   -e layer1-file2
+test   -e layer1-file3
+
+test   -e layer2-file1
+test ! -e layer2-file2
+test   -e layer2-file3
+
+test   -e layer3-file1
+test   -e layer3-file2
+test ! -e layer3-file3
+
+grep -q yes layer1-file2
+grep -q yes layer1-file3
+
+grep -q yes layer2-file1
+grep -q yes layer2-file3
+
+grep -q yes layer3-file1
+grep -q yes layer3-file2
+
+
+test   -e layer4-file1
+test ! -e layer4-file2
+test ! -e layer4-file3
+
+test ! -e layer5-file1
+test   -e layer5-file2
+test ! -e layer5-file3
+
+test ! -e layer6-file1
+test ! -e layer6-file2
+test   -e layer6-file3
+
+grep -q yes layer4-file1
+grep -q yes layer5-file2
+grep -q yes layer6-file3
+
+
+grep -q NEW layer10-file1
+grep -q NEW layer10-file2
+grep -q NEW layer10-file3
+
+grep -q line1 layer11-file1
+grep -q line1 layer11-file2
+grep -q line1 layer11-file3
+
+# # Docker with AUFS or overlay storage backend does not handle this test
+# # correctly and Semaphore uses AUFS
+if [ "$DOCKER_STORAGE_BACKEND" == devicemapper ] ; then
+	grep -q line2 layer11-file1
+	grep -q line2 layer11-file2
+	grep -q line2 layer11-file3
+	cmp layer11-file1 layer11-file2
+	cmp layer11-file1 layer11-file3
+fi
+
+echo "SUCCESS"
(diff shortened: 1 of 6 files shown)


tests/test.sh
```diff
@@ -25,19 +25,30 @@ if ! which rkt > /dev/null ; then
...
+	echo "### Test case ${TESTNAME}: test in Docker..."
+	sudo docker run --rm --env=CHECK=docker-run --env=DOCKER_STORAGE_BACKEND=$DOCKER_STORAGE_BACKEND $PREFIX/${TESTNAME}
+
+	echo "### Test case ${TESTNAME}: converting to ACI..."
 	sudo docker save -o ${TESTNAME}.docker $PREFIX/${TESTNAME}
 	$DOCKER2ACI ${TESTNAME}.docker
-	sudo $RKT run --insecure-skip-verify ./${PREFIX}-${TESTNAME}-latest.aci
+
+	echo "### Test case ${TESTNAME}: test in rkt..."
+	sudo $RKT run --insecure-skip-verify --set-env=CHECK=rkt-run --set-env=DOCKER_STORAGE_BACKEND=$DOCKER_STORAGE_BACKEND ./${PREFIX}-${TESTNAME}-latest.aci
```
iaguis: We should remove the pod we create (maybe with rkt prepare & rkt run-prepared & rkt rm $UUID) and the image from the CAS.

tests/test-whiteouts/check.sh
```diff
@@ -0,0 +1,66 @@
...
+grep -q NEW layer10-file2
+grep -q NEW layer10-file3
+
+grep -q line1 layer11-file1
+grep -q line1 layer11-file2
+grep -q line1 layer11-file3
+
+# # Docker with AUFS or overlay storage backend does not handle this test
+# # correctly and Semaphore uses AUFS
+if [ "$DOCKER_STORAGE_BACKEND" == devicemapper ] ; then
```
alban: I fixed this. This test works fine with Docker/devicemapper but not with Docker/aufs or Docker/overlayfs.

tests/test.sh
alban: Fixed.

tests/test.sh
```diff
@@ -25,19 +25,40 @@ if ! which rkt > /dev/null ; then
...
+
+	echo "### Test case ${TESTNAME}: test in rkt..."
+	sudo $RKT prepare --insecure-skip-verify \
+	                  --set-env=CHECK=rkt-run \
+	                  --set-env=DOCKER_STORAGE_BACKEND=$DOCKER_STORAGE_BACKEND \
+	                  ./${PREFIX}-${TESTNAME}-latest.aci \
+	                  > rkt-uuid-${TESTNAME}
+	sudo $RKT run-prepared $(cat rkt-uuid-${TESTNAME})
+	sudo $RKT status $(cat rkt-uuid-${TESTNAME}) | grep app-${TESTNAME}=0
+	sudo $RKT rm $(cat rkt-uuid-${TESTNAME})
```
iaguis: Cool, we still leave the images behind, that's annoying especially since every run they have a different hash. We should also remove them.

It would also be nice to run these clean commands if any part of the script exits (for example, if the exit status is not 0) but we can do that later.

tests/test-whiteouts/check.sh
iaguis: :+1:

tests/test.sh
alban: For deleting images, I filed https://github.com/coreos/rkt/issues/1779. Can we do it after https://github.com/coreos/rkt/issues/1779 gets fixed?

iaguis: ok

iaguis:
LGTM

PR #107, opened 2015-12-01
common: make manifests deterministic

iaguis:
Basically, we sort every golang map before serializing the Image
Manifest.

There's still a problem with PaxHeaders causing generated some ACIs to be
non-deterministic (https://github.com/appc/docker2aci/issues/89).


alban:
LGTM

PR #109, opened 2015-12-04
docker2aci: separate "http" and "accept any certificate" flags

iaguis:
Instead of having an --insecure flag that uses "http://" if it's
activated we add two flags:

- --insecure-skip-verify: accepts any certificate from the server but
  continues using TLS.
- --insecure-registry: uses a plain unencrypted HTTP registry.

Helps with https://github.com/coreos/rkt/issues/1829
Maps well with https://github.com/coreos/rkt/issues/1836

### DIFF
--- a/lib/backend/repository/repository.go
+++ b/lib/backend/repository/repository.go
@@ -20,23 +20,25 @@ import (
 )
 
 type RepositoryBackend struct {
-	repoData          *RepoData
-	username          string
-	password          string
-	insecure          bool
-	hostsV2Support    map[string]bool
-	hostsV2AuthTokens map[string]map[string]string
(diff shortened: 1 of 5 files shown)


jonboulle:
Looks OK. @dgonyeo ?

iaguis:
We're not sure if we should just keep one `insecure` with the following meaning:

* --insecure: accepts any certificate from the server, if the server doesn't support TLS, allow HTTP connections.

dgonyeo:
I think having two flags makes more sense. From a UX standpoint, maybe still have an `--insecure` flag that does the same thing as if you had provided both of these flags? Might be too frivolous though.

iaguis:
Closing in favor of #163

Issue #110, opened 2015-12-13
Allow selecting the compression method


kamalmarhubi:
The [ACI spec](https://github.com/appc/spec/blob/0875c991a496fe2b65ed296551c667b199a6bfb8/spec/aci.md#image-archives) says

> Image archives MAY be compressed with gzip, bzip2, or xz.

It would be handy if the docker2aci CLI could let the user choose the compression method.

kamalmarhubi:
I'm throwing together a PR that at least makes compression optional, adding a `--compression` flag. Neither bzip2 nor xz have compressors in the standard library, so I'll leave the allowed options as `none` and `gzip` (default).

iaguis:
Fixed via #111

PR #111, opened 2015-12-13
Allow disabling compression

kamalmarhubi:
This adds a `--compression` flag which can take the values `gzip` and
`none`. It is enum-like rather than boolean to allow xz and bzip2 to be
added later.

Refs #110

### DIFF
--- a/lib/common/common.go
+++ b/lib/common/common.go
@@ -33,6 +33,13 @@ const (
 	appcDockerParentImageID = "appc.io/docker/parentimageid"
 )
 
+type Compression int
+
+const (
+	NoCompression = iota
+	GzipCompression
+)
+
 func ParseDockerURL(arg string) *types.ParsedDockerURL {
 	if arg == "" {
 		return nil
@@ -59,7 +66,7 @@ func ParseDockerURL(arg string) *types.ParsedDockerURL {
 	}
 }
 
-func GenerateACI(layerNumber int, layerData types.DockerImageData, dockerURL *types.ParsedDockerURL, outputDir string, layerFile *os.File, curPwl []string, compress bool) (string, *schema.ImageManifest, error) {
+func GenerateACI(layerNumber int, layerData types.DockerImageData, dockerURL *types.ParsedDockerURL, outputDir string, layerFile *os.File, curPwl []string, compression Compression) (string, *schema.ImageManifest, error) {
 	manifest, err := GenerateManifest(layerData, dockerURL)
 	if err != nil {
 		return "", nil, fmt.Errorf("error generating the manifest: %v", err)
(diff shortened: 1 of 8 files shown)


lib/common/common.go
```diff
@@ -33,6 +33,13 @@ const (
 	appcDockerParentImageID = "appc.io/docker/parentimageid"
 )
 
+type Compression int
+
+const (
+	NoCompression = iota
+	GzipCompression = iota
```
iaguis: You don't need this `iota`

lib/docker2aci.go
```diff
@@ -224,10 +229,19 @@ func getManifests(renderedACI acirenderer.RenderedACI, aciRegistry acirenderer.A
...
+func writeSquashedImage(outputFile *os.File, renderedACI acirenderer.RenderedACI, aciProvider acirenderer.ACIProvider, manifests []schema.ImageManifest, compression common.Compression) error {
+	var tarWriterTarget io.WriteCloser = outputFile
+
+	switch(compression) {
+	case common.NoCompression:
+	case common.GzipCompression:
+		tarWriterTarget = gzip.NewWriter(outputFile)
+		defer tarWriterTarget.Close()
+	default:
+		return fmt.Errorf("Unexpected compression enum value: %d", compression)
```
iaguis: we should return lower-cased errors

lib/common/common.go
```diff
@@ -358,7 +365,7 @@ func writeACI(layer io.ReadSeeker, manifest schema.ImageManifest, curPwl []strin
 	defer aciFile.Close()
 
 	var w io.WriteCloser = aciFile
-	if compress {
+	if compression != NoCompression {
```
iaguis: Shouldn't we check specifically for gzip here?

iaguis:
Thanks for the PR! I've made some comments but looks good :)

kamalmarhubi:
@iaguis thanks for the quick response! As an operational note for addressing the comments, do you prefer commits to be added, or to rewrite the commit?

lib/docker2aci.go
kamalmarhubi: Done.

lib/common/common.go
kamalmarhubi: Cargo culting... I don't really know go! :p

Done.

lib/common/common.go
kamalmarhubi: Done.

kamalmarhubi:
I've pushed them as added commits. Let me know if you want me to squash before you merge!

iaguis:
Yeah, please squash them and then I'll merge it. Thanks a lot!

kamalmarhubi:
All done! Thanks again for the quick review. I'm tempted to put in xz and bzip2 compression (using installed utilities if available) base on that alone!

PR #113, opened 2016-01-22
backend/repository2: set auth headers for getLayer

iaguis:
It was causing authenticated docker fetches to fail.

### DIFF
--- a/lib/backend/repository/repository2.go
+++ b/lib/backend/repository/repository2.go
@@ -56,9 +56,7 @@ func (rb *RepositoryBackend) supportsV2(indexURL string) (bool, error) {
 		return false, err
 	}
 
-	if rb.username != "" && rb.password != "" {
-		req.SetBasicAuth(rb.username, rb.password)
-	}
+	rb.setBasicAuth(req)
 
 	client := &http.Client{}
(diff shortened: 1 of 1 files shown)


lib/backend/repository/repository2.go
```diff
@@ -198,6 +198,10 @@ func (rb *RepositoryBackend) getLayerV2(layerID string, dockerURL *types.ParsedD
 		return nil, err
 	}
 
+	if rb.username != "" && rb.password != "" {
+		req.SetBasicAuth(rb.username, rb.password)
+	}
+
```
alban: Is it the equivalent of `setAuthTokenV1` in `getLayerV1()`?

iaguis: Yes. IIRC, in docker v1 you authenticate against the index, receive a token and then use that to query the registry. In docker v2, there's no distinction between index and registry, so you always use authentication.

alban: This pattern is now used 4 times in this file. Maybe factorize:
```
rb.setBasicAuth(req)
```

jonboulle:
Lgtm if  it works!

PR #114, opened 2016-01-25
lib: allow relative paths in ImageManifest's App.Exec

iaguis:
Since https://github.com/appc/spec/pull/527 the spec allows relative
paths in App.Exec, so we don't need to prepend "/bin/sh -c".

### DIFF
--- a/lib/common/common.go
+++ b/lib/common/common.go
@@ -440,18 +440,7 @@ func writeACI(layer io.ReadSeeker, manifest schema.ImageManifest, curPwl []strin
 }
 
 func getExecCommand(entrypoint []string, cmd []string) appctypes.Exec {
-	var command []string
-	if entrypoint == nil && cmd == nil {
-		return nil
-	}
-	command = append(entrypoint, cmd...)
-	// non-absolute paths are not allowed, fallback to "/bin/sh -c command"
-	if len(command) > 0 && !filepath.IsAbs(command[0]) {
-		command_prefix := []string{"/bin/sh", "-c"}
-		quoted_command := util.Quote(command)
-		command = append(command_prefix, strings.Join(quoted_command, " "))
-	}
-	return command
+	return append(entrypoint, cmd...)
 }
 
 func parseDockerUser(dockerUser string) (string, string) {


alban:
LGTM if it works.

lib/common/common.go
```diff
@@ -440,18 +440,10 @@ func writeACI(layer io.ReadSeeker, manifest schema.ImageManifest, curPwl []strin
 }
 
 func getExecCommand(entrypoint []string, cmd []string) appctypes.Exec {
-	var command []string
 	if entrypoint == nil && cmd == nil {
```
iaguis: Hm, I pushed an old version, this is not needed. If both `entrypoint` and `cmd` are nil, `append(entrypoint, cmd...)` returns `nil` already.

alban:
ok, LGTM as well.

iaguis:
Tested with busybox ([it uses a relative path](https://github.com/docker-library/busybox/blob/81c593a39bec858e92bfa2c239aa57e7b988ba37/uclibc/Dockerfile)):

```
$ docker2aci docker://busybox
Downloading sha256:eeee0535bf3: [==============================] 676 KB/676 KB 
Downloading sha256:a3ed95caeb0: [==============================] 32 B/32 B

Generated ACI(s):
library-busybox-latest.aci
$ actool cat-manifest --pretty-print library-busybox-latest.aci
{
    "acKind": "ImageManifest",
    "acVersion": "0.7.0",
    "name": "registry-1.docker.io/library/busybox",
    "labels": [
        {
            "name": "version",
            "value": "latest"
        },
        {
            "name": "os",
            "value": "linux"
        },
        {
            "name": "arch",
            "value": "amd64"
        }
    ],
    "app": {
        "exec": [
            "sh"
        ],
        "user": "0",
        "group": "0"
    },
    "annotations": [
        {
            "name": "created",
            "value": "2016-01-15T18:06:41Z"
        },
        {
            "name": "appc.io/docker/registryurl",
            "value": "registry-1.docker.io"
        },
        {
            "name": "appc.io/docker/repository",
            "value": "library/busybox"
        },
        {
            "name": "appc.io/docker/imageid",
            "value": "964092b7f3e54185d3f425880be0b022bfc9a706701390e0ceab527c84dea3e3"
        },
        {
            "name": "appc.io/docker/parentimageid",
            "value": "9e77fef7a1c9f989988c06620dabc4020c607885b959a2cbd7c2283c91da3e33"
        }
    ]
}
$ sudo rkt --insecure-options=image run --interactive library-busybox-latest.aci
rkt: using image from file /home/iaguis/work/coreos/go/src/github.com/coreos/rkt/build-rkt/bin/stage1-coreos.aci
rkt: using image from file library-busybox-latest.aci
/ # exit
```

lib/common/common.go
```diff
@@ -440,18 +440,7 @@ func writeACI(layer io.ReadSeeker, manifest schema.ImageManifest, curPwl []strin
 }
 
 func getExecCommand(entrypoint []string, cmd []string) appctypes.Exec {
```
jonboulle: can just inline this now..

PR #117, opened 2016-01-27
lib/backend/repository: handle http and https gracefully

iaguis:
If docker2aci is passed the `--insecure` flag, don't directly use http but
try https first and then http.

Helps with https://github.com/coreos/rkt/issues/2036

### DIFF
--- a/lib/backend/repository/repository.go
+++ b/lib/backend/repository/repository.go
@@ -14,9 +14,20 @@
 package repository
 
 import (
+	"fmt"
+	"net/http"
+	"path"
+
 	"github.com/appc/docker2aci/lib/common"
 	"github.com/appc/docker2aci/lib/types"
 	"github.com/appc/spec/schema"
+)
+
+type registryVersion int
+
+const (
+	registryV1 registryVersion = iota
+	registryV2
 )
 
 type RepositoryBackend struct {
@@ -26,6 +37,7 @@ type RepositoryBackend struct {
 	insecure          bool
 	hostsV2Support    map[string]bool
 	hostsV2AuthTokens map[string]map[string]string
+	schema            string
 	imageManifests    map[types.ParsedDockerURL]v2Manifest
 }
 
@@ -44,18 +56,29 @@ func (rb *RepositoryBackend) GetImageInfo(url string) ([]string, *types.ParsedDo
 	dockerURL := common.ParseDockerURL(url)
 
 	var supportsV2, ok bool
+	var URLSchema string
 	if supportsV2, ok = rb.hostsV2Support[dockerURL.IndexURL]; !ok {
 		var err error
-		supportsV2, err = rb.supportsV2(dockerURL.IndexURL)
+		URLSchema, supportsV2, err = rb.supportsRegistry(dockerURL.IndexURL, registryV2)
 		if err != nil {
 			return nil, nil, err
 		}
+		rb.schema = URLSchema + "://"
+		supportsV2 = false
 		rb.hostsV2Support[dockerURL.IndexURL] = supportsV2
 	}
 
 	if supportsV2 {
 		return rb.getImageInfoV2(dockerURL)
 	} else {
+		URLSchema, supportsV1, err := rb.supportsRegistry(dockerURL.IndexURL, registryV1)
+		if err != nil {
+			return nil, nil, err
+		}
+		if !supportsV1 {
+			return nil, nil, fmt.Errorf("registry doesn't support API v2 nor v1")
+		}
+		rb.schema = URLSchema + "://"
 		return rb.getImageInfoV1(dockerURL)
 	}
 }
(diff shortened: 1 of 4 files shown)


iaguis:
Tested with:

* Official Docker registry with and without `--insecure`
 * Works
* Local HTTP registry without `--insecure`:
 * `Error: conversion error: Get https://localhost:5000/v2: tls: oversized record received with length 20527`
* Local HTTP registry with `--insecure`:
 * Works
* quay with forced v1 API with and without `--insecure`:
 * Works

lib/backend/repository/repository.go
```diff
@@ -44,18 +49,28 @@ func (rb *RepositoryBackend) GetImageInfo(url string) ([]string, *types.ParsedDo
 	dockerURL := common.ParseDockerURL(url)
 
 	var supportsV2, ok bool
+	var URLSchema string
 	if supportsV2, ok = rb.hostsV2Support[dockerURL.IndexURL]; !ok {
 		var err error
-		supportsV2, err = rb.supportsV2(dockerURL.IndexURL)
+		URLSchema, supportsV2, err = rb.supportsRegistry(dockerURL.IndexURL, "v2")
```
krnowak: It isn't my turf, but I would make registry version an enum.

lib/backend/repository/repository.go
```diff
@@ -68,10 +83,66 @@ func (rb *RepositoryBackend) BuildACI(layerNumber int, layerID string, dockerURL
...
+		url := schema + "://" + URLStr
+		req, err := http.NewRequest("GET", url, nil)
+		if err != nil {
+			return nil, err
+		}
+
+		rb.setBasicAuth(req)
+
+		client := &http.Client{}
+		res, err = client.Do(req)
```
krnowak: Maybe just use `http.Do(req)`?

lib/backend/repository/repository.go
```diff
@@ -68,10 +83,66 @@ func (rb *RepositoryBackend) BuildACI(layerNumber int, layerID string, dockerURL
...
+		return
+	}
+
+	schema = "https"
+	res, err := fetch(schema)
+	if err == nil {
+		ok, err = checkRegistryStatus(res.StatusCode, res.Header, version)
```
krnowak: I would do `defer res.Body.Close()` here.

lib/backend/repository/repository.go
```diff
@@ -68,10 +83,66 @@ func (rb *RepositoryBackend) BuildACI(layerNumber int, layerID string, dockerURL
...
+	}
+	if err != nil || !ok {
+		if rb.insecure {
+			if res != nil {
+				res.Body.Close()
+			}
+			schema = "http"
+			res, err = fetch(schema)
+			if err == nil {
+				ok, err = checkRegistryStatus(res.StatusCode, res.Header, version)
```
krnowak: Same here.

lib/backend/repository/repository1.go
```diff
@@ -94,7 +94,7 @@ func (rb *RepositoryBackend) buildACIV1(layerNumber int, layerID string, dockerU
 
 func (rb *RepositoryBackend) getRepoDataV1(indexURL string, remote string) (*RepoData, error) {
 	client := &http.Client{}
-	repositoryURL := rb.protocol() + path.Join(indexURL, "v1", "repositories", remote, "images")
```
krnowak: Should this commit remove the `protocol` function then?

lib/backend/repository/repository.go
iaguis: ok

lib/backend/repository/repository.go
iaguis: there's no `http.Do()`, there's `http.Get()` but it doesn't allow messing with the headers.

lib/backend/repository/repository1.go
iaguis: Ooops... forgot about that

iaguis: Hmm... wait, it *is* removed in this commit

iaguis:
Updated

lib/backend/repository/repository.go
krnowak: Ah, indeed. Nevermind then.

lib/backend/repository/repository1.go
krnowak: Ooops… ETOOMANYREVIEWS.

lib/backend/repository/repository.go
```diff
@@ -14,11 +14,22 @@
...
+
 	"github.com/appc/docker2aci/lib/common"
 	"github.com/appc/docker2aci/lib/types"
 	"github.com/appc/spec/schema"
 )
 
+type registryVersion int
+
+const (
+	registryV1 registryVersion = 1 << iota
```
krnowak: Just `iota` is enough, you bit-maniac. ;)

iaguis: But I like shifting!

lib/backend/repository/repository.go
```diff
@@ -68,10 +91,67 @@ func (rb *RepositoryBackend) BuildACI(layerNumber int, layerID string, dockerURL
...
+
+	schema = "https"
+	res, err := fetch(schema)
+	if err == nil {
+		ok, err = checkRegistryStatus(res.StatusCode, res.Header, version)
+		defer res.Body.Close()
+	}
+	if err != nil || !ok {
+		if rb.insecure {
+			if res != nil {
```
krnowak: This is not needed here, since you added the defer above.

iaguis: oops... missed that one :disappointed:

krnowak:
Alright, LFAD from a bystander. ;)

iaguis:
Thanks!

PR #118, opened 2016-01-29
lib/repository: do not force Docker API v1

iaguis:
I added this statement to force using Docker API v1 and test if it
worked and I forgot to remove it before pushing :$.


alban:
LGTM

PR #119, opened 2016-02-08
lib/common: return error instead of nil if docker url is empty

iaguis:


main.go
```diff
@@ -77,6 +77,7 @@ func runDocker2ACI(arg string, flagNoSquash bool, flagImage string, flagDebug bo
 
 		aciLayerPaths, err = docker2aci.Convert(dockerURL, squash, ".", os.TempDir(), compression, username, password, flagInsecure)
 	} else {
+		fmt.Println(arg)
```
krnowak: That's an accidental debugging leftover I think.

krnowak:
Looks ok, but I was surprised that the only situation when you can get an error is to pass an empty string. Looks like `23%4kjh3KJ#H$3h43jh' is a valid docker url. ;)

main.go
iaguis: Du hast Recht

iaguis:
> Looks ok, but I was surprised that the only situation when you can get an error is to pass an empty string. Looks like `23%4kjh3KJ#H$3h43jh' is a valid docker url. ;)

Sure:

```
> docker2aci 'docker://23%4kjh3KJ#H$3h43jh'
Error: error parsing argument: parse docker://23%4kjh3KJ: percent-encoded characters in host
[oops, URL parsing error]
> docker2aci 'docker://234kjh3KJ#H$3h43jh'
Error: conversion error: unexpected http code: 404, URL: https://registry-1.docker.io/v2/library/234kjh3KJ#H$3h43jh/manifests/latest
```

However:

```
> docker pull '23%4kjh3KJ#H$3h43jh'
Using default tag: latest
repository name component must match "[a-z0-9](?:-*[a-z0-9])*(?:[._][a-z0-9](?:-*[a-z0-9])*)*"
```

I guess we could match that regexp too, but should we bother?

krnowak:
> I guess we could match that regexp too, but should we bother?

Probably not, unless we get a bug report about some string that is that looks like a docker URL, but it isn't it.

jonboulle:
surely we can just anticipate that and be a little proactive here?

iaguis:
Matching that regexp is not enough because that's only the hostname part. 

We'll have to adapt the contents of this package (except the digest part, since we're not using it (yet?)) https://github.com/docker/distribution/tree/master/reference

lib/backend/file/file.go
```diff
@@ -29,7 +29,12 @@ func NewFileBackend(file *os.File) *FileBackend {
 }
 
 func (lb *FileBackend) GetImageInfo(dockerURL string) ([]string, *types.ParsedDockerURL, error) {
-	parsedDockerURL := common.ParseDockerURL(dockerURL)
+	fmt.Println(dockerURL)
```
alban: debugging leftover

iaguis:
Updated

lib/common/common.go
```diff
@@ -40,9 +40,13 @@ const (
 	GzipCompression
 )
 
-func ParseDockerURL(arg string) *types.ParsedDockerURL {
+func ParseDockerURL(arg string) (*types.ParsedDockerURL, error) {
 	if arg == "" {
-		return nil
+		return nil, fmt.Errorf("empty Docker image reference")
```
krnowak: There is nothing to replace in the format string, so maybe use `errors.New`?

lib/common/common.go
```diff
@@ -40,9 +40,13 @@ const (
...
+	}
+
+	if !referenceRegexp.MatchString(arg) {
+		return nil, fmt.Errorf("invalid Docker image reference")
```
krnowak: Maybe make the invalid reference a part of the message.

krnowak:
Minor nits, otherwise LFAD. I only skimmed over the regexp thing.

iaguis:
Updated.

krnowak:
Alright, LFAD.

PR #120, opened 2016-02-19
lib/common: infer default docker registry

iaguis:
If the Docker reference does not include any servers, we assume it's a
using the default Docker registry.

Since the Docker API v1 was deprecated in the default registry, we can
now infer the default docker registry name (registry-1.docker.io).

Needed to fix https://github.com/coreos/rkt/issues/2008

### DIFF
--- a/lib/common/common.go
+++ b/lib/common/common.go
@@ -51,12 +51,12 @@ func ParseDockerURL(arg string) *types.ParsedDockerURL {
 	}
 	indexURL, imageName := SplitReposName(taglessRemote)
 
-	if indexURL == "" && !strings.Contains(imageName, "/") {
+	// the Docker client considers images referenced only by a name (e.g.
+	// "busybox" or "ubuntu") as valid, and, in that case, it adds the
+	// "library/" prefix because that's how they're stored in the official
+	// registry
+	if indexURL == defaultIndexURL && !strings.Contains(imageName, "/") {
 		imageName = "library/" + imageName
-	}
-
-	if indexURL == "" {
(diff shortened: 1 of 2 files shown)


lib/common/common.go
```diff
@@ -51,14 +51,10 @@ func ParseDockerURL(arg string) *types.ParsedDockerURL {
 	}
 	indexURL, imageName := SplitReposName(taglessRemote)
 
-	if indexURL == "" && !strings.Contains(imageName, "/") {
+	if indexURL == defaultIndexURL && !strings.Contains(imageName, "/") {
```
krnowak: Is this something specific to the registry at docker.io, but not to the one at, say, localhost?

krnowak:
One question, otherwise LFAD.

lib/common/common.go
iaguis: Yep, the Docker client considers images referenced only by a name (e.g. `busybox`, `ubuntu` instead of `coreos/etcd` or `localhost/whatever`) as valid, and, in that case, it adds the `library/` prefix because that's how they're stored in the docker.io registry.

krnowak: That's… stupid. It's like leaking implementation detail on all the docker clients. Meh.

Anyway, thanks for making it clear. Maybe a comment would be useful here, because it quite not obvious.

iaguis: Alright. BTW, here's the code in Docker: https://github.com/docker/docker/blob/117a982d2e805d3279ff75fcec071635e5191ef6/reference/reference.go#L171

PR #124, opened 2016-03-03
lib/common: make a copy of pwl instead of reusing the same var

iaguis:
In f78ecc1d6e7c3fd913b0d374f759a5b6d3a628f2, we started sorting every
golang map before serializing the Image Manifests to make the generated
ACIs deterministic (golang randomizes maps).

This causes problems with squashed ACIs: they're missing some files (see
https://github.com/coreos/rkt/issues/2244 for some examples).

The reason for this is that we were modifying the PathWhitelist inside
writeACI() and then reusing the same copy in the next iteration of the
loop in convertReal(). Making a copy of the modified PathWhitelist when
we generate the layer's manifest fixes the problem.

cc @krnowak

### DIFF
--- a/tests/rkt-v1.1.0.md5sum
+++ b/tests/rkt-v1.1.0.md5sum
@@ -1,0 +1,1 @@
+MD5 (rkt-v1.1.0.tar.gz) = d3d9d62429e53d8f631dbec93e4e719f
--- a/lib/common/common.go
+++ b/lib/common/common.go
@@ -433,10 +433,14 @@ func writeACI(layer io.ReadSeeker, manifest schema.ImageManifest, curPwl []strin
 	}
 	newPwl := subtractWhiteouts(curPwl, whiteouts)
 
-	manifest.PathWhitelist, err = writeStdioSymlinks(trw, fileMap, newPwl)
+	newPwl, err = writeStdioSymlinks(trw, fileMap, newPwl)
 	if err != nil {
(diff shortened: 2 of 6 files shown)


krnowak:
"lib/common: don't sort the PathWhitelists" isn't correct.

jonboulle:
can we test?

krnowak:
@jonboulle: Maybe. But I don't know what happens exactly (if it is defined) if you do something like this for a large value in`X` and large array in returned by `gimmeSomeStrings()`:
```go
var list []string
var lol [][]string
for i := 0; i < X; i++ {
    list = append(list, gimmeSomeStrings()...)
    lol = append(lol, list)
}
```

iaguis:
Added a stupid test that converts one of the failing images and checks that `/var/run` is present.

tests/rkt-v1.1.0.md5sum
```diff
@@ -0,0 +1 @@
+MD5 (rkt-v1.1.0.tar.gz) = d3d9d62429e53d8f631dbec93e4e719f
```
jonboulle: md5 is so 1992

jonboulle:
lgtm

tests/rkt-v1.1.0.md5sum
iaguis: I blame @alban, he's oldschool

PR #125, opened 2016-03-05
build: use `/usr/bin/env bash` instead of `/bin/bash`

dgonyeo:
Using `/usr/bin/env bash` instead of `/bin/bash` makes the build script
a little more portable, as it's not always safe to assume that
`/bin/bash` exists (for example, this is the case in NixOS).

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,4 +1,6 @@
-#!/bin/bash -e
+#!/usr/bin/env bash
+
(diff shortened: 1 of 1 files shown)


iaguis:
Huh, I forgot we had a `build.sh`.

Happy to help with



LGTM

PR #126, opened 2016-03-05
lib/backend/repository; handle when auth challenge is missing scope

dgonyeo:
On some registries (like quay) the scope portion of the bearer auth
challenge can be omitted. In these situations docker2aci has all the
pieces it needs to infer the scope of the token to request, so this
commit changes docker2aci to do so when the scope is missing.


iaguis:
I don't have a private quay account to test but I assume you did and it works.

LGTM.

PR #129, opened 2016-03-07
*: prepare release

iaguis:
This PR includes commits that prepare the first docker2aci release:

* Clean the API by reducing the exposed functions and using config structs instead of individual parameters.
* Decouple an library error message from the docker2aci CLI
* Change the names of the Convert functions
* Exit with status 2 when the arguments are invalid
* Add version flag
* Update build instructions

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -4,6 +4,9 @@ set -e
 
 ORG_PATH="github.com/appc"
 REPO_PATH="${ORG_PATH}/docker2aci"
+# TODO: use "git describe --dirty" when we actually have tags pushed
+VERSION=$(git rev-parse --short HEAD)
+GLDFLAGS="-X github.com/appc/docker2aci/lib.Version=${VERSION}"
 
 if [ ! -h gopath/src/${REPO_PATH} ]; then
 	mkdir -p gopath/src/${ORG_PATH}
@@ -18,5 +21,4 @@ eval $(go env)
 echo "Fetching dependencies..."
 go get -d -v ./...
 echo "Building docker2aci..."
-go build -o $GOBIN/docker2aci ${REPO_PATH}/
-
+go build -o $GOBIN/docker2aci -ldflags "${GLDFLAGS}" ${REPO_PATH}/
--- a/lib/version.go
+++ b/lib/version.go
@@ -1,0 +1,20 @@
+// Copyright 2015 The appc Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package docker2aci
+
+import "github.com/appc/spec/schema"
+
+var Version = "0.0.0+was-not-built-correctly"
+var AppcVersion = schema.AppContainerVersion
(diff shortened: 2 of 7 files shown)


build.sh
```diff
@@ -4,6 +4,9 @@ set -e
 
 ORG_PATH="github.com/appc"
 REPO_PATH="${ORG_PATH}/docker2aci"
+# TODO: use "git describe --dirty" when we actually have tags pushed
+VERSION=$(git rev-parse --short HEAD)
+GLDFLAGS="-X github.com/appc/docker2aci/lib.Version=${VERSION}"
```
krnowak: That won't work with go1.4. ;)

krnowak: Also, when vendoring docker2aci, the `Version` variable in `lib` package will not have the correct version (it will be `0.0.0+was-not-built-correctly` as hardcoded there), because we cannot expect project vendoring docker2aci code to pass a necessary linker flag with correct version.

A possible workaround could be hardcoding the version 0.9.0 and use some bump-version script, rkt style.

This would make docker2aci gogettable.

lib/version.go
```diff
@@ -0,0 +1,20 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package docker2aci
+
+import "github.com/appc/spec/schema"
+
+var Version = "0.0.0+was-not-built-correctly"
```
krnowak: If you get rid of the linker flag in favor of presumably the bump-version script, then you can make both `Version` and `AppcVersion` const.

main.go
```diff
@@ -35,9 +36,15 @@ var (
...
+	flagVersion     = flag.Bool("version", false, "Print version")
 )
 
-func runDocker2ACI(arg string, flagNoSquash bool, flagImage string, flagDebug bool, flagInsecure bool, flagCompression string) error {
+func printVersion() {
+	fmt.Println("docker2aci version", docker2aci.Version)
+	fmt.Println("appc version", docker2aci.AppcVersion)
+}
+
+func runDocker2ACI(arg string, flagNoSquash bool, flagImage string, flagDebug bool, flagInsecure bool, flagCompression string, flagVersion bool) error {
```
krnowak: I guess you really didn't want to use `*flagImage` or `*flagNoSquash` everytime. Maybe instead of shadowing global variables with local ones you could simply do:
```go
var (
    flagNoSquash bool = false
    flagCompression string = "gzip"
    …
)

func init() {
    flag.BoolVar(&flagNoSquash, "nosquash", "Blurb here")
    flag.StringVar(&flagCompression, "compression", "Blah blah blah")
    …
}
```
This way you can have the `runDocker2ACI` function with just one parameter.

krnowak: Besides, passing a `flagVersion` parameter here is not necessary, you don't use it here at all.

iaguis: Yep, thanks!

build.sh
iaguis: I don't think we need to care about go1.4

How about I set the default version in `lib/version.go` to `0.9.0` and `0.9.0+git` like we do in rkt (with a bump-version script), but if you happen to build with `./build.sh` you also get the git commit with `git describe`?

lib/version.go
krnowak: Talked about it with @iaguis OOB - it might be better to move these variables into docker2aci main package instead if we want to keep the way docker2aci is built.

build.sh
krnowak: Yeah, I don't mind if docker2aci drops support for go1.4.

About version - not sure if it is best of both worlds or worst of both worlds. But that will work too, yes.

krnowak:
The version stuff can be reworked in some followup PR. LFAD.

PR #130, opened 2016-03-08
lib,tarball: document missing exported functions

iaguis:
Also, make license headers consistent.


tarball/tarfile.go
```diff
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// Packege tarball provides functions to manipulate tar files.
```
krnowak: Package

iaguis: Ouch

krnowak:
One typo spotted. I'm wondering if documentation of the `docker2aci` subpackages (`types`, `backend` and `util`) should mention that they are an implementation detail and should not be used. Dunno about `common` subpackage - it exports `Compression` enumeration which is used in the exported `docker2aci.CommonConfig` type.

lib/backend/file/file.go
```diff
@@ -1,3 +1,19 @@
+// Copyright 2015 The appc Authors
```
alban: 2016?

iaguis: Welp, the code was written in 2015...

iaguis:
> One typo spotted. I'm wondering if documentation of the docker2aci subpackages (types, backend and util) should mention that they are an implementation detail and should not be used. Dunno about common subpackage - it exports Compression enumeration which is used in the exported docker2aci.CommonConfig type.

I'll add a note for types, backend and util.

iaguis:
Updated.

lib/common/common.go
```diff
@@ -44,6 +59,17 @@ const (
 	GzipCompression
 )
 
+type ErrSeveralImages struct {
```
krnowak: Docs missing.

krnowak:
Consider splitting the `common` package into two - the exported `common` package and the `internal` package being an implementation detail.

tarball/tarfile.go
```diff
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// Package tarball provides functions to manipulate tar files.
```
krnowak: Mark it as implementation detail - if somebody complains, we can probably export it as a `pkg` subpackage?

iaguis: Alright

iaguis:
> Consider splitting the common package into two - the exported common package and the internal package being an implementation detail.

OK, I'll do that

iaguis:
Updated.

lib/docker2aci.go
```diff
@@ -37,11 +38,21 @@ import (
 	appctypes "github.com/appc/spec/schema/types"
 )
 
+// Docker2ACIBackend is the interface that abstracts converting Docker layers
```
krnowak: This can be moved to the `internal` package now.

krnowak:
Hopefully last nit.

iaguis:
Hopefully...

krnowak:
LFAD.

iaguis:
\ø/

PR #131, opened 2016-03-08
release v0.9.0

iaguis:


krnowak:
LFAD. The `git describe` will work when there is a tag.

PR #134, opened 2016-03-09
build.sh: remove fetching dependencies

iaguis:
We're vendoring the dependencies now, there's no need to fetch them.


krnowak:
Tested with go1.{4,5,6}. LFAD.

Issue #133, opened 2016-03-08
Export annotation keys (e.g. appcDockerEntrypoint, appcDockerCmd, etc)


yifan-gu:
During image conversion, some information are recorded as [image annotations](https://github.com/appc/docker2aci/blob/v0.9.0/lib/internal/internal.go#L50-L56).
We might want to export there annotations keys so that users can reference them from other projects.

cc @iaguis

PR #135, opened 2016-03-09
lib: move internal packages to lib/internal and export ParseDockerURL and annotations

iaguis:
Good things brought by this PR:

* Internal functions are clearly separated from the rest
* People can use annotations from outside docker2aci
* We weren't exporting ParseDockerURL which is used by rkt, fix that

Fixes https://github.com/appc/docker2aci/issues/133


lib/common/common.go
```diff
@@ -30,3 +45,9 @@ type ErrSeveralImages struct {
 func (e *ErrSeveralImages) Error() string {
 	return e.Msg
 }
+
+// ParseDockerURL takes a Docker URL and returns a ParsedDockerURL with its
+// index URL, image name, and tag.
+func ParseDockerURL(arg string) (*types.ParsedDockerURL, error) {
```
krnowak: This is troublesome - the returned result is a part of internal package, so the user of the `ParseDockerURL` function can use it directly, but cannot write a function that takes the `*types.ParsedDockerURL`, because the user will not be able to import the internal package…

I think that putting the `ParsedDockerURL` struct in the `common` package will end up in cyclic dependencies again, so consider adding `type ParsedDockerURL types.ParsedDockerURL` somewhere in the `common` package and do some casting in the `ParseDockerURL` function.

krnowak:
Something to consider before releasing 1.0 in a separate PR - replace the `pkg/log` with a `Logger` field in the `CommonConfig` struct, so users can pass their own logger. The `Logger` fields would be of some interface type like:

```go
type Logger interface {
    Debug(i ...interface{})
    Info(i ...interface{})
}
```
so the implementer could implement the `Debug` function in the same way as currently `pkg/log` does without resorting to some global boolean variable set by the `InitDebug()` function. Possibly export two implementation like `DefaultLogger` with configurable debugging function and `DevNullLogger`.

krnowak:
Alright, I think I'll put my comment into a separate issue.

LFAD.

krnowak:
Filed #136.

iaguis:
Thanks!

PR #140, opened 2016-03-16
lib: use upstream functions and parse digest

iaguis:
Now that we vendor, we can get rid of some copied code.

Also, this might help with https://github.com/coreos/rkt/issues/2188

cc @dgonyeo

### DIFF
--- a/lib/internal/docker/docker_functions.go
+++ b/lib/internal/docker/docker_functions.go
@@ -17,16 +17,16 @@ package docker
 import (
 	"encoding/base64"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io/ioutil"
 	"os"
 	"path"
-	"regexp"
 	"runtime"
 	"strings"
 
 	"github.com/appc/docker2aci/lib/internal/types"
+
+	"github.com/docker/distribution/reference"
 )
 
 const (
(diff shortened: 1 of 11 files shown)


lib/internal/docker/docker_functions.go
```diff
@@ -35,118 +35,21 @@ const (
...
-		remoteName = reposName
+// SplitReposName breaks a repo name into an index name and remote name.
+func SplitReposName(name string) (indexName, remoteName string) {
+	i := strings.IndexRune(name, '/')
+	if i == -1 || (!strings.ContainsAny(name[:i], ".:") && name[:i] != "localhost") {
+		indexName, remoteName = defaultIndexURL, name
 	} else {
-		indexName = nameParts[0]
-		remoteName = nameParts[1]
+		indexName, remoteName = name[:i], name[i+1:]
```
krnowak: This case will be visited also when `i` is `-1`, so you cannot use it for indexing, right?

krnowak: This case will be visited also when `i` is `-1`, so you cannot use it for indexing, right?

iaguis: if `i` is `-1` it will fall under the `if` condition, oder?

krnowak: Eh, nevermind.

lib/internal/docker/docker_functions.go
```diff
@@ -232,31 +135,27 @@ func GetAuthInfo(indexServer string) (string, string, error) {
...
-		return nil, fmt.Errorf("invalid Docker image reference %q", arg)
+	r, err := reference.ParseNamed(arg)
+	if err != nil {
+		return nil, err
 	}
 
-	taglessRemote, tag := parseRepositoryTag(arg)
-	if tag == "" {
-		tag = defaultTag
+	tag := "latest"
```
krnowak: `tag := defaultTag`?

iaguis: yep

krnowak:
One nit. Otherwise LFAD.

PR #142, opened 2016-03-17
change docker entrypoint/cmd annotations to json

sjpotter:
can't serialize them with a join as exec arguments can have a space

### DIFF
--- a/lib/internal/internal.go
+++ b/lib/internal/internal.go
@@ -182,10 +182,18 @@ func GenerateManifest(layerData types.DockerImageData, dockerURL *types.ParsedDo
 
 	if dockerConfig != nil {
 		if len(dockerConfig.Entrypoint) > 0 {
-			annotations = append(annotations, appctypes.Annotation{Name: *appctypes.MustACIdentifier(common.AppcDockerEntrypoint), Value: strings.Join(dockerConfig.Entrypoint, " ")})
+			entry, err := json.Marshal(dockerConfig.Entrypoint)
+			if err != nil {
+				return nil, err
+			}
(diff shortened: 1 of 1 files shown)


sjpotter:
@yifan-gu

yifan-gu:
LGTM

yifan-gu:
@iaguis

iaguis:
Can you prefix the commit with "lib/internal:"?

This is just so the users can inspect how the original image was, right?

iaguis:
> This is just so the users can inspect how the original image was, right?

I'll answer myself: yes (I just checked the original PR for these annotations).

LGTM after changing the commit message

lib/internal/internal.go
```diff
@@ -182,10 +182,18 @@ func GenerateManifest(layerData types.DockerImageData, dockerURL *types.ParsedDo
...
+                       }
+                       annotations = append(annotations, appctypes.Annotation{Name: *appctypes.MustACIdentifier(common.AppcDockerEntrypoint), Value: string(entry)})
 		}
 		if len(dockerConfig.Cmd) > 0 {
-			annotations = append(annotations, appctypes.Annotation{Name: *appctypes.MustACIdentifier(common.AppcDockerCmd), Value: strings.Join(dockerConfig.Cmd, " ")})
+                       cmd, err := json.Marshal(dockerConfig.Cmd)
+                       if err != nil {
+                               return nil, err
+                       }
+                       annotations = append(annotations, appctypes.Annotation{Name: *appctypes.MustACIdentifier(common.AppcDockerCmd), Value: string(cmd)})
```
iaguis: Weird, I wonder why `Value` is colored here and not above...

yifan-gu: @sjpotter I think you didn't gofmt it.

PR #150, opened 2016-04-02
Fix auth token parse to handle service with space.

choppsv1:
Fix for case where service has a space in it (e.g., service="Docker regsitry").


dgonyeo:
LGTM

PR #155, opened 2016-05-12
Optionally use insecure client also for v1 Docker repositories

### LINKED ISSUE #144: error certificate signed by unknown authority while docker pull ok
Hi,
I try to convert an image from a private repository based on Docker registry v2.
A docker pull works fine, but when executing against docker2aci I have a *certificate signed by unknown authority* error.

    # bin/docker2aci docker://docker-registry.genouest.org/osallou/testgit
    Error: conversion error: Get https://docker-registry.genouest.org/v2: x509: certificate signed by unknown authority

Via Docker:

    # docker pull docker-registry.genouest.org/osallou/testgit
    latest: Pulling from docker-registry.genouest.org/osallou/testgit
    1986b9e4a6af: Already exists
    ....

jsoriano:
Fixes #144 for docker registry v1.

### DIFF
--- a/lib/internal/backend/repository/repository1.go
+++ b/lib/internal/backend/repository/repository1.go
@@ -29,6 +29,7 @@ import (
 	"github.com/appc/docker2aci/lib/common"
 	"github.com/appc/docker2aci/lib/internal"
 	"github.com/appc/docker2aci/lib/internal/types"
+	"github.com/appc/docker2aci/lib/internal/util"
 	"github.com/appc/docker2aci/pkg/log"
 	"github.com/appc/spec/schema"
 	"github.com/coreos/ioprogress"
@@ -95,7 +96,7 @@ func (rb *RepositoryBackend) buildACIV1(layerNumber int, layerID string, dockerU
 }
 
 func (rb *RepositoryBackend) getRepoDataV1(indexURL string, remote string) (*RepoData, error) {
-	client := &http.Client{}
+	client := util.GetTLSClient(rb.insecure)
 	repositoryURL := rb.schema + path.Join(indexURL, "v1", "repositories", remote, "images")
 
 	req, err := http.NewRequest("GET", repositoryURL, nil)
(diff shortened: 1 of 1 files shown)


dgonyeo:
I believe `rb.insecure` gets set by the `--insecure` flag, which is described in the help page as "Uses unencrypted connections when fetching images".

Not using an encrypted connection and not verifying the TLS certificate on the encrypted connection are two different things, so I think either the help output should be updated to reflect that it allows both, or an additional flag should be added for this.

iaguis:
> or an additional flag should be added for this.

I agree.

iaguis:
https://github.com/appc/docker2aci/pull/109?

jsoriano:
I agree too with having two different flags, the problem is that since 89c6437b3b45ec53d7af5ce5920932d88bb666e9 pulling from docker registries with self-signed certificates is not possible, and it affects rkt since ~1.2. It worked in previous versions.
This PR just tried to solve this regression by using the `util.GetTLS` function as it's being done in v2 registry.
Any solution for this would be welcomed :)

iaguis:
What about doing what we do in rkt and having these two config options in `RemoteConfig`:

```
InsecureSkipTLS   bool
InsecureAllowHTTP bool
```

Then we'd have two different flags in the docker2aci CLI and in rkt we'd call `docker2aci.ConvertRemoteRepo()` with the corresponding values according to the rkt options `--insecure-options=http` and `--insecure-options=tls`.

jsoriano:
The two flags approach looks the way to go, yes, was there any problem with #109?

iaguis:
It's been a while and I don't really remember. It has to be reworked anyway, things have changed.

jsoriano:
@iaguis take a look to https://github.com/jsoriano/docker2aci/tree/split-insecure, shall I create another PR with this?

iaguis:
sounds good, thanks!

PR #161, opened 2016-05-25
Added missing message to channel on succesful layer download

jsoriano:
Without this it hangs forever when layers are succesfully downloaded.


dgonyeo:
Well that's embarrassing. Thanks!

PR #168, opened 2016-05-30
lib/repository2: use new progressutil interface

alban:
This bumps github.com/coreos/pkg/progressutil to v2, fixing the races from https://github.com/coreos/pkg/pull/64.


lucab:
LGTM

s-urbaniak:
LGTM

PR #176, opened 2016-06-10
lib/repository2: add support for docker v2.2 and OCI

dgonyeo:
This commit adds support for the docker v2.2 image format and OCI image
format. When the remote registry supports the docker v2 HTTP API,
docker2aci will add the media types for v2.1, v2.2 and OCI to the Accept
header on the manifest request, and handle the returned manifest version
accordingly.


dgonyeo:
Right now docker2aci is relying on the docker hub rewriting manifests into the v2.1 schema for it.

Also I want to write a test or two for this before it gets merged, so that will come next week.

jonboulle:
@iaguis @s-urbaniak review?

lib/internal/backend/repository/repository2.go
```diff
@@ -167,48 +175,146 @@ func (rb *RepositoryBackend) buildACIV2(layerIDs []string, dockerURL *types.Pars
...
+
+	tmpParentDir, err := ioutil.TempDir(tmpBaseDir, "docker2aci-")
+	if err != nil {
+		return nil, nil, err
+	}
+	defer os.RemoveAll(tmpParentDir)
+
+	copier := progressutil.NewCopyProgressPrinter()
+
+	var errChannels []chan error
```
s-urbaniak: any reason why we cannot use a single (buffered) channel). The amount of err's is known in forehand.

lib/internal/backend/repository/repository2.go
```diff
@@ -167,48 +175,146 @@ func (rb *RepositoryBackend) buildACIV2(layerIDs []string, dockerURL *types.Pars
...
+	closers := make([]io.ReadCloser, len(layerIDs))
+	for i, layerID := range layerIDs {
+		errChan := make(chan error, 1)
```
s-urbaniak: see above, you could create a single `errChan := make(chan error, len(layerIDs))`

lib/internal/backend/repository/repository2.go
```diff
@@ -167,48 +175,146 @@ func (rb *RepositoryBackend) buildACIV2(layerIDs []string, dockerURL *types.Pars
...
+		go func() {
+			tmpDir, err := ioutil.TempDir(tmpParentDir, "")
+			if err != nil {
+				errChan <- fmt.Errorf("error creating dir: %v", err)
+				return
+			}
+
+			layerFiles[i], closers[i], err = rb.getLayerV2(layerID, dockerURL, tmpDir, copier)
+			if err != nil {
+				errChan <- fmt.Errorf("error getting the remote layer: %v", err)
```
s-urbaniak: you are using `errChan` as the sync mechanism to propagate changes of `layerFiles[i]`, and `closers[i]`, this is pretty fragile. Wouldn't it be easier to simply send a `layer struct` via a channel it being:
```
type layer struct {
  file *os.File
  c io.ReadCloser
  err error
}
```

dgonyeo:
@s-urbaniak switched to using a single buffered channel, as per your suggestion

s-urbaniak:
@dgonyeo that was quick ;-) thanks for the quick update. How did you test this locally? I'd like to play with it tomorrow morning.

dgonyeo:
I pointed it at the docker hub (i.e. `./bin/docker2aci docker://nginx`) and used print statements to see that my new code paths were being hit. Not graceful, but works.

For the tests I'm going to write this week I intend to spin up a local [distribution](https://github.com/docker/distribution), but docker2aci gets upset when the local distribution responds to its https request with http, so I might need to tweak `--insecure-allow-http` to support that.

lib/internal/backend/repository/repository2.go
```diff
@@ -167,48 +175,156 @@ func (rb *RepositoryBackend) buildACIV2(layerIDs []string, dockerURL *types.Pars
...
+				index:  i,
+				file:   layerFile,
+				closer: closer,
+				err:    nil,
+			}
+		}()
+	}
+	for i := 0; i < len(layerIDs); i++ {
+		res := <-resultChan
+		defer res.closer.Close()
```
s-urbaniak: this will panic if `res.closer` is nil which can happen if you return an `err`.

s-urbaniak: I guess we can just swap with the err check below to prevent this.

lib/internal/backend/repository/repository2.go
```diff
@@ -167,48 +175,156 @@ func (rb *RepositoryBackend) buildACIV2(layerIDs []string, dockerURL *types.Pars
...
+		if res.err != nil {
```
s-urbaniak: If an error occurs at `i`, we might end up with response bodies not closed for all layers with `i < index < len(layerIDs)`

dgonyeo:
@s-urbaniak how's it look now:

- a buffered channel is used to receive results from the goroutines started up
- the closers and files, if not nil, have a deferred close (realized we could also leave the files open here if an error was hit a few lines down)
- If the err is not nil, it's put in a list
- After getting the responses from each goroutine, if the list of errors is not empty, return the first error that was encountered

s-urbaniak:
@dgonyeo :+1: ... those dreaded resources ;-)

Another idea I had to omit those `if closer/file != nil` checks: we could also return a https://golang.org/pkg/io/ioutil/#NopCloser in case of an err .. but I leave that to you if that makes code any "prettier".

lgtm from my side, didn't make it to test today though, will do so tomorrow.

s-urbaniak:
@dgonyeo out of curiosity: why do we need to bubble up those response bodies (closer) up through the channel? Can't we close them earlier?

dgonyeo:
@s-urbaniak: those response bodies can't be closed until the `copier.PrintAndWait` call finishes, as the `CopyProgressPrinter` is still copying data from them. Maybe it would be cleaner to have the `CopyProgressPrinter` have a function like `CloseAllInputs` that could be called after the transfers are complete.

jonboulle:
bump - what's left? more review?

s-urbaniak:
Still lgtm from my side, just had a small implementation question.

dgonyeo:
Just realized that I never correctly added the annotations to the new manifest after doing the v2.2 -> AppC manifest conversion, so just updated this commit to fix that.

dgonyeo:
And just added a commit that tweaks the tests to run them against a local registry in addition to running them on local files. The tests aren't as comprehensive as I wanted, but I couldn't find an easy way to generate/seed the registry with differing versions of docker images. This at least hits the fetching logic a few times.

jonboulle:
@s-urbaniak final review?

lib/internal/backend/repository/repository2.go
```diff
@@ -167,7 +175,106 @@ func (rb *RepositoryBackend) buildACIV2(layerIDs []string, dockerURL *types.Pars
...
+		if res.closer != nil {
+			defer res.closer.Close()
+		}
+		if res.file != nil {
+			defer res.file.Close()
+		}
+		if res.err != nil {
+			errs = append(errs, res.err)
+		}
+		layerFiles[res.index] = res.file
```
s-urbaniak: `res.file` can be `nil` in case of an err, hence `layerFiles[res.index]` will be nil too.

lib/internal/backend/repository/repository2.go
```diff
@@ -167,7 +175,106 @@ func (rb *RepositoryBackend) buildACIV2(layerIDs []string, dockerURL *types.Pars
...
+	}
+	if len(errs) > 0 {
+		return nil, nil, errs[0]
+	}
+	err = copier.PrintAndWait(os.Stderr, 500*time.Millisecond, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+	for _, layerFile := range layerFiles {
+		err := layerFile.Sync()
```
s-urbaniak: so this will panic if `layerFiles[res.index]` is nil, see above.

s-urbaniak:
Found two small nits

lib/internal/backend/repository/repository2.go
dgonyeo: In the case where there's an error, it'll be appended to `errs`, which gets checked [here](https://github.com/appc/docker2aci/pull/176#discussion-diff-67860822R241), so by the time it gets to this line every element in `layerFiles` should be valid.

s-urbaniak: ah, got, overlooked that line ;-) thanks, lgtm then

lib/internal/backend/repository/repository2.go
jonboulle: seems like this is taken care of, res always has index now

lib/internal/backend/repository/repository2.go
```diff
@@ -167,7 +175,106 @@ func (rb *RepositoryBackend) buildACIV2(layerIDs []string, dockerURL *types.Pars
...
+			resultChan <- layer{
+				index:  i,
+				file:   layerFile,
+				closer: closer,
+				err:    nil,
+			}
+		}()
+	}
+	var errs []error
+	for i := 0; i < len(layerIDs); i++ {
```
jonboulle: for res := range resultChan

lib/internal/internal.go
```diff
@@ -248,6 +276,133 @@ func GenerateManifest(layerData types.DockerImageData, dockerURL *types.ParsedDo
 	return genManifest, nil
 }
 
+func GenerateEmptyManifest(name string) (*schema.ImageManifest, error) {
+	acid, err := appctypes.NewACIdentifier(name)
+	if err != nil {
+		return nil, err
+	}
+
+	archlabel, err := appctypes.NewACIdentifier("arch")
```
jonboulle: MustACIdentifier

lib/internal/internal.go
```diff
@@ -248,6 +276,133 @@ func GenerateManifest(layerData types.DockerImageData, dockerURL *types.ParsedDo
...
+	if err != nil {
+		return nil, err
+	}
+
+	oslabel, err := appctypes.NewACIdentifier("os")
```
jonboulle: MustACIdentifier

jonboulle:
Couple of nits but also LGTM overall

lib/internal/backend/repository/repository2.go
dgonyeo: Cool, I didn't know you could range over channels. But `resultChan` never gets closed, so wouldn't the for loop never exit?

dgonyeo:
Oh, also added a gitignore because things in `test/` were cluttering my `git status`.

lib/internal/backend/repository/repository2.go
s-urbaniak: yes, if you don't close the channel, range will block forever.

s-urbaniak:
@dgonyeo ping, is it ready for merge?

dgonyeo:
I believe it is.

jonboulle:
@dgonyeo @s-urbaniak

s-urbaniak:
Sorry for the delay, I was confused who is supposed to merge.

PR #180, opened 2016-07-07
avoid O(N) and fix defer reader close

runcom:
- First commit deals with an O(N) routine and removes it (the last layer ID index takes N-1 cycle to be found)
- Second commit remove a reader close call because the reader is used after the function returns

### DIFF
--- a/lib/internal/backend/repository/repository2.go
+++ b/lib/internal/backend/repository/repository2.go
@@ -102,9 +102,9 @@ func (rb *RepositoryBackend) buildACIV21(layerIDs []string, dockerURL *types.Par
 
 			manifest := rb.imageManifests[*dockerURL]
 
-			layerIndex, err := getLayerIndex(layerID, manifest)
-			if err != nil {
-				errChan <- err
+			layerIndex, ok := rb.reverseLayers[layerID]
+			if !ok {
+				errChan <- fmt.Errorf("layer not found in manifest: %s", layerID)
 				return
 			}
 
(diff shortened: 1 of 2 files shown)


lib/internal/backend/repository/repository2.go
```diff
@@ -512,7 +504,6 @@ func (rb *RepositoryBackend) getLayerV2(layerID string, dockerURL *types.ParsedD
 			if err != nil {
 				return nil, nil, err
 			}
-			defer res.Body.Close()
```
dgonyeo: I just noticed that the [original res](https://github.com/appc/docker2aci/pull/180/files#diff-acab445bd3e69965398509002dd14a18L499) doesn't get closed if this if block gets dropped into. Would you mind adding a `res.Body.Close()` around like 507 while you're here?

dgonyeo:
LGTM

lib/internal/backend/repository/repository2.go
runcom: there's a problem however, response body closing is an issue in all of this routine, I'll add a defer

runcom:
@dgonyeo reworked the second commit so that if there's an error and the `res` var isn't nil then we close the body

lib/internal/backend/repository/repository2.go
```diff
@@ -496,23 +492,32 @@ func (rb *RepositoryBackend) getLayerV2(layerID string, dockerURL *types.ParsedD
...
 	}
 
-	res, err := rb.makeRequest(req, dockerURL.ImageName, accepting)
+	res, err = rb.makeRequest(req, dockerURL.ImageName, accepting)
 	if err != nil {
 		return nil, nil, err
 	}
 
+	defer func() {
+		if err != nil {
```
iaguis: Just "and" those to save indentation?

iaguis:
One nit but LGTM afterwards.

lib/internal/backend/repository/repository2.go
runcom: changed to and

iaguis:
Thanks!

PR #181, opened 2016-07-08
tests: added golang tests to lib/tests to test image fetching

dgonyeo:
This commit creates the directory lib/tests, which includes the
following:

 - code for generating a manifest and layers conforming to docker v2.2
 - code for emulating a docker registry that hosts these images (the
   docker daemon can successfully pull images from this)
 - a few tests that utilize these to test using the docker2aci library
   to pull/convert an image

These fetching tests triggered a panic in docker2aci, which this commit also fixes.

I intend to create a followup PR to also test the docker v2.1 logic and the logic for converting a file on disk, and to hopefully pull the logic under tests/ into these golang tests.

### DIFF
--- a/lib/tests/common.go
+++ b/lib/tests/common.go
@@ -1,0 +1,138 @@
+package test
+
+import (
+	"archive/tar"
+	"bytes"
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
+	"io/ioutil"
+	"os"
+	"path"
+
+	"github.com/appc/docker2aci/lib/internal/typesV2"
+)
+
+type Layer map[*tar.Header][]byte
+
+type Docker22Image struct {
+	RepoTags []string
+	Layers   []Layer
+	Config   typesV2.ImageConfig
+}
+
+func GenerateDocker22(destPath string, img Docker22Image) error {
+	layerHashes, err := GenLayers(destPath, img.Layers)
+	if err != nil {
+		return err
+	}
+	configHash, err := GenDocker22Config(destPath, img.Config, layerHashes)
+	if err != nil {
+		return err
+	}
+	err = GenDocker22Manifest(destPath, configHash, layerHashes)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func GenLayers(destPath string, layers []Layer) ([]string, error) {
+	var layerHashes []string
+	for _, l := range layers {
+		layerBuffer := &bytes.Buffer{}
+		tw := tar.NewWriter(layerBuffer)
+		for hdr, contents := range l {
+			hdr.Size = int64(len(contents))
+			err := tw.WriteHeader(hdr)
+			if err != nil {
+				tw.Close()
+				return nil, err
+			}
+			_, err = tw.Write(contents)
+			if err != nil {
+				tw.Close()
+				return nil, err
+			}
+		}
+		tw.Close()
+		layerTarBlob := layerBuffer.Bytes()
+		h := sha256.New()
+		h.Write(layerTarBlob)
+		hashStr := hex.EncodeToString(h.Sum(nil))
+		layerHashes = append(layerHashes, hashStr)
+		err := ioutil.WriteFile(path.Join(destPath, hashStr), layerTarBlob, 0644)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return layerHashes, nil
+}
+
+func GenDocker22Config(destPath string, conf typesV2.ImageConfig, layerHashes []string) (string, error) {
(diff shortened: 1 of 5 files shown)


lib/tests/common.go
```diff
@@ -0,0 +1,144 @@
...
+	"encoding/hex"
+	"encoding/json"
+	"io/ioutil"
+	"os"
+	"path"
+
+	"github.com/appc/docker2aci/lib/internal/typesV2"
+)
+
+type Layer map[*tar.Header][]byte
```
s-urbaniak: This will use tar.Header pointer values as keys, not the tar.Header values. If put two equal values of `tar.Header`, but with distinct memory locations, they will have a different key.

iaguis:
We should run these tests on Semaphore

jonboulle:
or Jenkins?

iaguis:
> or Jenkins?

:trollface:

jonboulle:
not trolling :-(

iaguis:
What's the benefit of running this on several distros?

jonboulle:
I don't think we need to do it on several different distros - but it would
be nice to move to a point where we don't need semaphore

iaguis:
> I don't think we need to do it on several different distros - but it would
> be nice to move to a point where we don't need semaphore

I don't see a pressing need for this, I think Semaphore works fine for docker2aci's use case.

lib/tests/common.go
dgonyeo: When I make that change I get:

```
derek@proton ~/go/src/github.com/appc/docker2aci> go test -v github.com/appc/docker2aci/lib/tests
# github.com/appc/docker2aci/lib/tests                                                          
lib/tests/common.go:16: invalid map key type tar.Header
```

No clue why go doesn't allow this, but I don't think that using pointers will be an issue here. This is only used in the creation of tests, and test writers can avoid creating duplicate `tar.Header`s.

dgonyeo:
Added these tests to `tests/test.sh`, and they now get run by semaphore.

lib/tests/common.go
s-urbaniak: Ah, I overlooked you're trying to use stdlib's `tar.Header` as the key type. http://godoc.org/archive/tar#Header contains `Xattrs` of type `map[string]string` and that is the problem.

Map [key types](https://golang.org/ref/spec#KeyType) must have the comparison operators == and != fully defined, hence your struct key type must not contain a function, map, or slice.

I agree that this is used only in tests, but if you'll ever put two equal header values, but with different memory locations, they won't be the same :-/

s-urbaniak: Suggestion: can't you take the layer hashes as the map keys?

dgonyeo: What I'm doing here is trying to work around the fact that Go doesn't have tuples. I want to store a list of pairs, where each pair is a tar header and the file contents. Go's maps kinda let me do this, by being a way to define a set of pairs (where each key and its value is the pair).

I could specify a struct that holds a `tar.Header` and a `[]byte`, and have a layer be a slice of these structs, but that also would allow duplicate tar headers, so there's no benefit there over the current approach.

I don't think using layer hashes as the map keys here makes sense, as I don't know the hashes at this point in time and the keys here represent a given file in the layer, not the layer itself.

To truly prevent duplicates perhaps I could go with the slice of structs approach and have a helper function to search the slice each time to make sure that I'm not adding duplicates, but that seems rather heavy handed for hardcoded tests.

s-urbaniak: You are right about the layer hashes, it is too late to do reviews ;-) Let's leave it like this then.

If I'll come up with a better idea, I'll do a follow-up PR.

LGTM

lib/tests/common.go
```diff
@@ -0,0 +1,144 @@
...
+
+type Docker22Image struct {
+	RepoTags []string
+	Layers   []Layer
+	Config   typesV2.ImageConfig
+}
+
+func GenerateDocker22(destPath string, img Docker22Image) error {
+	tmpDir, err := ioutil.TempDir("", "docker2aci-test-")
```
iaguis: This doesn't seem to be used.

iaguis:
The tests don't check whether the converted image has the right contents, right?

I guess that's tested on the tests in `tests/`.

dgonyeo:
That's right, at this point these tests aim only to exercise the fetching logic. I'd like to move the checks under `tests/` into these golang tests, but I figured it would be best to get this in before messing around with those for the sake of the docker2aci bump in rkt.

iaguis:
OK. Please remove the unused code I pointed out above and I'll merge it :)

dgonyeo:
Removed.

PR #183, opened 2016-07-17
added code for using rkt to build the project

rbucker:
I want to be able to produce an executable for CoreOS host (static binary) and to be able to build the project with rkt as the only requirement.


iaguis:
I see you're also changing the build script to build a statically compiled docker2aci. You use `-tags=netgo` which means it will use golang's dns resolver instead of the one in glibc. Although simple `/etc/resolv.conf` options work, this means the features not supported by go's resolver are not gonna work.

I'm hesitant about this, do you really need a statically compiled docker2aci? I understand that to build with glibc you'll need to switch alpine (which uses musl) to something that uses it.

For the record, we tried to get a static rkt but we dropped it (https://github.com/coreos/rkt/pull/1380) although things seem to work better nowadays

rbucker:
I do not have a particular opinion other than effort has been expended in this direction based on experience.

rbucker:
ok, so I have a small opinion now and in no particular order... [a] needs to run on CoreOS and it appears glibc is not included [b] docker2aci is licensed under the apache license and statically linking to glibc will virulently attach the GPL and that entire argument chain [c] even the official golang docker container uses Alpine. Clearly perfection is not possible so then something that is good enough needs to be good enough.

iaguis:
I'd rather keep it dynamic by default, maybe you can add a build-option to build statically?

Also, what about using our new [rkt-builder](https://github.com/coreos/rkt-builder), it has everything we need and it is trusted?

This should just work:

```
sudo rkt run \
    --volume src-dir,kind=host,source=${SRC_DIR} \
    --volume build-dir,kind=host,source=${SRC_DIR} \
    --interactive \
    coreos.com/rkt/builder:1.0.1 \
    --exec /opt/rkt/build.sh
```

README.md
```diff
@@ -18,6 +18,18 @@ disabled by specifying `--compression=none`.
 	cd docker2aci
 	./build.sh
 
+## Build with rkt
+
+assuming that you have rkt installed on you system
```
iaguis: Assuming that you have rkt installed on your system, this command will build docker2aci using it. The resulting binary will be in the `bin` folder .

iaguis:
> needs to run on CoreOS and it appears glibc is not included

It is included as far as I can see:

```
core@core-01 ~ $ ldd ./docker2aci 
	linux-vdso.so.1 (0x00007ffdacbfd000)
	libpthread.so.0 => /lib64/libpthread.so.0 (0x00007f593c61b000)
	libc.so.6 => /lib64/libc.so.6 (0x00007f593c270000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f593c838000)
```

rbucker:
On Mon, Jul 18, 2016 at 9:39 AM, Iago López Galeiras <
notifications@github.com> wrote:

> It is included as far as I can see


I was looking for the wrong filename

*--Richard Bucker *

*http://about.me/richard.bucker <http://about.me/richard.bucker>*

README.md
iaguis: Can you change the text to what I wrote on the comment.

build.sh
```diff
@@ -30,4 +30,5 @@ if [ "${GOOS}" = "freebsd" ]; then
 fi
 
 echo "Building docker2aci..."
-go build -o ${GOBIN}/docker2aci -ldflags "${GLDFLAGS}" ${REPO_PATH}
+CGO_ENABLED=0 go build -o ${GOBIN}/docker2aci -a -tags netgo -ldflags "${GLDFLAGS}" ${REPO_PATH}
```
iaguis: Please revert the static building.

iaguis:
Can you squash the commits and change the commit message to our [conventional format](https://github.com/coreos/rkt/blob/master/CONTRIBUTING.md#format-of-the-commit-message) (I should document it in this repo too). Something like:

```
build: add support for building with rkt
```

After that and after my inline comments are addressed we can merge this.

jonboulle:
bump?

lucab:
Closing a stale PR. This can be restarted at a later point, though.

PR #186, opened 2016-07-28
tests: remove redundant code and simplify code

iaguis:
We forgot to remove it when refactoring it to a function.

Since we're touching this, let's simplify the code a bit.


alban:
LGTM

PR #188, opened 2016-07-28
repository2: get the correct layer index

iaguis:
In da56c93f we introduced an optimization to avoid going through all the
layers every time we need the index of a specific one.

However, this resulted in incorrect Image Manifests being generated.

In f8e82fab we attempted to fix this. We assumed that the mistake was
that the index we used was supposed to be reversed (from the end instead
of from the beginning of the layer list). This assumption was wrong. and
was probably motivated by the name of the map (`reverseLayers`).

The problem was that there can be several layers with the same BlobSum
(layerID) but different manifest and when we populated the map, we were
storing the last one for a given BlobSum. This is not what we were doing
previously, instead we were returning the first match.

Fix this by storing the first appearance of a BlobSum. Also, change the
name of `reverseLayers` to `layersIndex`.

- [ ] Test

### DIFF
--- a/lib/internal/backend/repository/repository.go
+++ b/lib/internal/backend/repository/repository.go
@@ -50,7 +50,7 @@ type RepositoryBackend struct {
 	imageManifests    map[types.ParsedDockerURL]v2Manifest
 	imageV2Manifests  map[types.ParsedDockerURL]*typesV2.ImageManifest
 	imageConfigs      map[types.ParsedDockerURL]*typesV2.ImageConfig
-	reverseLayers     map[string]int
+	layersIndex       map[string]int
 }
 
 func NewRepositoryBackend(username string, password string, insecure common.InsecureConfig) *RepositoryBackend {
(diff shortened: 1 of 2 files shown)


iaguis:
This is hard to test because it only happens with API v2.1 and not with v2.2 and I'm not really sure how to reproduce it. The only image that shows this behavior I know of is `quay.io/stackanetes/stackanetes-memcached:barcelona`

iaguis:
@dgonyeo can you have a look?

Maybe you can even think of a way to test this...

dgonyeo:
I've been wanting to add emulation of a v2.1 registry to the new golang tests I added (https://github.com/appc/docker2aci/commit/b173254c583b41d9d0bfc6f0bb6d84d9223c1e93), but this would take some time to implement, and I don't expect to have any spare time for at least a few weeks. 

The changes would probably be:
- adding support for generating v2.1 images
- modifying the registry code to be able to serve v2.1 images
- adding tests using this code to test the v2.1 codepath

iaguis:
I think I'll merge this without a test for now then.

PR #199, opened 2016-09-14
lib/internal: add entrypoint/cmd annotations to v21 images

### LINKED ISSUE #198: Missing ENTRYPOINT and CMD annotations in manifest with newer Docker versions
Hi @yifan-gu , @jonboulle 

Related to: https://github.com/appc/docker2aci/issues/122 , @mlin and I noticed that there appears to be an issue where the manifest file is missing ENTRYPOINT and CMD annotations when newer versions of Docker are used to build images and those images are subsequently posted to Dockerhub.

An example to reproduce this issue is:

Using two versions of Docker (in our case we used 1.6.2 and 1.11.2, the current versions on Ubuntu 14.04 and 16.04 respectively):

- Build a simple Docker image using both versions of Docker:

```
echo "FROM ubuntu:14.04
ENTRYPOINT ls" | docker build -
```

- Push each image to Dockerhub (in our example we have geetduggal/docker2aci-entrypoint-issue and dnamlin/docker2aci-entrypoint-issue built with Docker versions 1.6.2 and 1.11.2 respectively)

- Run `docker2aci docker://<repo>/<imagename>` on each of these images

We noticed that the manifest file for image associated with dnamlin (Docker version 1.11.2) did not contain the entrypoint annotation while the image associated with geetduggal (Docker version 1.6.2) does.

Consequently, we cannot emulate the ENTRYPOINT + CMD user experience as discussed in the original issue.

If we manually pull and save the image from Dockerhub (using `docker pull` and `docker save`), then the annotation is preserved.   Therefore it appears that something in the interaction between docker2aci and Dockerhub is causing this missing annotation and that's about as far as we have narrowed it down to.

iaguis:
Also, refactor some code.

And add a test.

Fixes #198

### DIFF
--- a/lib/internal/internal.go
+++ b/lib/internal/internal.go
@@ -146,6 +146,26 @@ func generateACIPath(outputDir, imageName, digest, tag, osString, arch string, l
 	return path.Join(outputDir, aciPath)
 }
 
+func generateEPCmdAnnotation(dockerEP, dockerCmd []string) (string, string, error) {
+	var entrypointAnnotation, cmdAnnotation string
+	if len(dockerEP) > 0 {
+		entry, err := json.Marshal(dockerEP)
+		if err != nil {
+			return "", "", err
+		}
+		entrypointAnnotation = string(entry)
+	}
+	if len(dockerCmd) > 0 {
+		cmd, err := json.Marshal(dockerCmd)
+		if err != nil {
+			return "", "", err
+		}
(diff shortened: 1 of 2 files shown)


jonboulle:
LGTM if it works

iaguis:
welp, there's a shiny new test that passes (and doesn't if I revert the fix).

jonboulle:
:sheep:

PR #200, opened 2016-09-23
Add support for converting OCI tarfiles

vishvananda:
This also switches to the new OCI media types.

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -40,6 +40,10 @@ imports:
   version: 19b0b332c9e4516a6370a0456e6182c3b5036720
 - name: github.com/klauspost/pgzip
   version: 95e8170c5d4da28db9c64dfc9ec3138ea4466fd4
+- name: github.com/opencontainers/image-spec
+  version: b472d5f4067ddca5cfbfea148dfaf7b42d43c5f0
+  subpackages:
+  - specs-go
 - name: github.com/spf13/pflag
   version: 94e98a55fb412fcbcfc302555cb990f5e1590627
 - name: go4.org
--- a/lib/internal/backend/file/file.go
+++ b/lib/internal/backend/file/file.go
@@ -28,14 +28,17 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"strings"
 
 	"github.com/appc/docker2aci/lib/common"
 	"github.com/appc/docker2aci/lib/internal"
 	"github.com/appc/docker2aci/lib/internal/docker"
 	"github.com/appc/docker2aci/lib/internal/tarball"
 	"github.com/appc/docker2aci/lib/internal/types"
+	"github.com/appc/docker2aci/lib/internal/typesV2"
 	"github.com/appc/docker2aci/pkg/log"
 	"github.com/appc/spec/schema"
+	spec "github.com/opencontainers/image-spec/specs-go"
 )
 
 type FileBackend struct {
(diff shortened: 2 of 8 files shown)


lib/internal/backend/file/file.go
```diff
@@ -28,23 +28,28 @@ import (
...
+	"github.com/appc/docker2aci/lib/internal/typesV2"
 	"github.com/appc/docker2aci/pkg/log"
 	"github.com/appc/spec/schema"
+	spec "github.com/opencontainers/image-spec/specs-go"
 )
 
 type FileBackend struct {
-	file *os.File
+	file  *os.File
+	fname string
```
jonboulle: does file.Name() work for this?

jonboulle:
Cool!
@dgonyeo do you have any bandwidth for a review?

vishvananda:
ah good call. I didn't realize the go file object stored the name. Changed it to use that.

dgonyeo:
I could take a look on Wednesday, scrambling to do a project for school currently

philips:
ping @dgonyeo

dgonyeo:
Wow, so sorry about dropping this. Reviewing now.

lib/internal/backend/file/file.go
```diff
@@ -112,19 +128,82 @@ func (lb *FileBackend) BuildACI(layerIDs []string, dockerURL *types.ParsedDocker
 	return aciLayerPaths, aciManifests, nil
 }
 
-func getImageID(file *os.File, dockerURL *types.ParsedDockerURL) (string, *types.ParsedDockerURL, error) {
+func (lb *FileBackend) BuildACIV22(layerIDs []string, dockerURL *types.ParsedDockerURL, outputDir string, tmpBaseDir string, compression common.Compression) ([]string, []*schema.ImageManifest, error) {
+	if len(layerIDs) < 2 {
+		return nil, nil, fmt.Errorf("insufficient layers for oci image")
```
dgonyeo: Could you explain this a little? Are single layer OCI images not possible?

vishvananda: see the above comment "// for oci the first image is the config"

To build an oci we need the sha of the image config object as well as the sha of the layers since it isn't in a known location like it is with v1 docker images. The sha of the image.json is being passed as the first item in the layers slice. It is pulled out below. This was the easiest way to get the data here without modifying the signature of a bunch of methods.

vishvananda:
rebased. Will try rkt tests

lib/internal/backend/file/file.go
dgonyeo: Ah gotcha, makes sense.

glide.lock
```diff
@@ -1,37 +1,37 @@
-hash: d088a750a1a81b463b0056b81edd3ce849b9bef0eb44861f67041e13b95bade0
-updated: 2016-07-06T16:33:39.979049165+02:00
+hash: bb06d4eb9b26ddb217ba46a9c1924a870e9abbf3259dabffd19fa4407b7d769c
+updated: 2016-10-17T14:22:17.272063593-07:00
 imports:
 - name: github.com/appc/spec
-  version: fd5e3d8f5d596c6fbd506e80b5e13ba8a97b704c
+  version: f329150233aa3ac6c83ab5ba02a2b5a8bd03d365
```
lucab: I see we are using tilde-versions in glide.yaml, so everything is getting bumped to latest patch release, but with a major-version of 0 this doesn't make sense. Do we perhaps want to pin down to exact versions instead?

/cc @s-urbaniak @jonboulle

vishvananda:
The unit tests pass fine with the new version. The functional tests are getting a lot of failures but i suspect they fail without my changes as well

glide.lock
vishvananda: glide doesn't seem to offer the ability to just update a single package. If there is a way to do this, I'm happy to just add the new package without touching anything else.

s-urbaniak: I am also not aware of a functionality to update just a single package in glide, unfortunately. That is why we pin versions in rkt to avoid general bumps like this. I suggest to pin versions and bump just the single package, or make a local test-run with unchecked tilde-versions in glide.yaml to check if there are note-worthy updates in other packages too.

vishvananda:
Ok I manually updated the glide.yaml and glide.lock and used glide install. The hash on the lock file doesn't match anymore, but this means that the update can be done separately after the merge.

dgonyeo:
Sorry about the couple day delay, was fighting with my setup over something dumb.

LGTM

Issue #203, opened 2016-09-30
Infinite loop vulnerability in retrieving images chain


TheBeeMan:
Hi,

In code reviewing, i found an infinite loop vulnerability in retrieving images chain using docker2aci, it occurs during the corresponding json file parsing from user's image archive, fetching the parent image ID until ID is nil. There must be a possibility that the images chain may be a closed cycle, thus , docker2aci will fall into an infinite loop, that's indeed true by some interesting tests.

I think the core cause of this issue is lacking in essential check for duplicated image ID, such as the current image ID could not be equal to its parent image ID, most important, check whether the images chain is a closed cycle.

I processed some interesting test for this issue, building a crafted image whose top layer's parent ID points to itself, then an infinite loop occurred, this flaw caused excessive CPU cycles & resources consume   on the host.

expecting subsequent discuss and fix the issue together, and could you request a CVE identifier for that ?

lucab:
Thanks for the report @TheBeeMan. From the description above, even if not explicitly stated it looks like your code review is covering the logic in [getAncestry](https://github.com/appc/docker2aci/blob/v0.12.3/lib/internal/backend/file/file.go#L288-L294) (please pinpoint if this covers something else instead).

In the above code, I agree that the current parent-handling behavior is buggy as it doesn't take into account cyclic dependencies.

However I'm not completely sure about any security implication of this. Triggering this infinite loop seems to require local access and only concern a single user process. This will just result in a never-ending conversion of a single malformed image, which shouldn't impact availability per-se as there is no single service to DoS here. As you pointed out, there will be indeed some over-usage of processing resources but this is something that is typically controlled by the OS (eg. via cgroups, priority-throttling, OOM-killing).

What's was your original impact analysis of this suggesting that it requires a CVE?

TheBeeMan:
This is a typical denial of service vulnerability, which could result in over-usage of processing resources. I think it is a CVE worthy vulnerability, although it is a low severity issue but a crafted image could cause docker2aci fall into a never-ending conversion.

The way you think about security is different with me, denial of service is a grey area, Obviously if I send 10 gigabits of request traffic or an valid input files and the processor gets slow/non responsive the CVE response would be "No CVE for you" ,but if a single crafted file prevents the whole system from working in an expected manner, that may be a problem that is worth a CVE.

Just for reference, infinite loop is a well-known security issue[](https://cwe.mitre.org/data/definitions/835.html)

TheBeeMan:
[...]

Issue #201, opened 2016-09-27
Path traversals present in  image converting


TheBeeMan:
### Description

**> in code reviewing, i found a path traversal vulnerability in docker's image converting using docker2aci, there must be a possibility that it extracts embedded layer data to arbitrary directories or paths since no essential  check for file path, RCE or privilege escalation would be performed.**

**> it is indeed true that i tested  the issue by building a malicious image, if running as root,  arbitrary file could be written into arbitrary paths, like backdoors, or running as unprivileged user,  arbitrary files also could be extracted to some paths within the capabilities of current user.**

**> It is quite critical,  right ? Could you request a CVE for that ?**

jonboulle:
Thanks for submitting this bug report. docker2aci development is primarily handled by CoreOS, so we'd like to handle this via our Security Disclosure policy. Could you kindly send an email to <security@coreos.com> with more details so we can investigate further?

https://coreos.com/security/disclosure/

TheBeeMan:
fine

TheBeeMan:
the issue has been sent to security@coreos.com by email, no response. I gonna report it to oss-security for handling, and request a CVE identifier, ok ?

lucab:
For [reference](http://www.openwall.com/lists/oss-security/2016/09/28/8), this has been assigned CVE-2016-7569 with a low to medium impact, typically mitigated for remote attack vectors.

Given the very short timeline for the disclosure, a patch is currently being worked on and will appear in the next release.

lucab:
@TheBeeMan a proposed fix for this is up at https://github.com/appc/docker2aci/pull/204, adding additional validation on crafted images. Can you please take a look at it?

TheBeeMan:
i reviewed the patch forCVE-2016-7569 and processed some tests with the previous malicious image, it addressed the issue.

PR #204, opened 2016-10-10
Additional validation on malformed images

lucab:
This PR introduces some additional validation before converting Docker images. This addresses:
 * an infinite loop when converting images with a cyclic dependency chain (CVE-2016-8579)
 * a path traversal when extracting layers with crafted IDs (CVE-2016-7569)

It also introduces fixtures and tests script for such cases.

Fixes https://github.com/appc/docker2aci/issues/201
Fixes https://github.com/appc/docker2aci/issues/203


lib/internal/backend/file/file.go
```diff
@@ -279,14 +282,24 @@ func extractEmbeddedLayer(file *os.File, layerID string, outputPath string) (*os
...
+// chain and errors out in such cases.
 func getAncestry(file *os.File, imgID string) ([]string, error) {
 	var ancestry []string
+	deps := make(map[string]bool)
 
 	curImgID := imgID
 
 	var err error
 	for curImgID != "" {
+		if _, ok := deps[curImgID]; ok {
```
jonboulle: nit, since you're using a bool you can just use the default value check (`if deps[curImgID]`)

lucab:
@jonboulle rebased to address the nit and to include the reference to the second CVE.

/cc @TheBeeMan

lib/common/common.go
```diff
@@ -60,3 +63,21 @@ func ParseDockerURL(arg string) (*ParsedDockerURL, error) {
 	p, err := docker.ParseDockerURL(arg)
 	return (*ParsedDockerURL)(p), err
 }
+
+// IsValidLayerRefs validates a layer refs
```
jonboulle: unused - remove this?

lib/internal/backend/file/file.go
```diff
@@ -279,14 +282,24 @@ func extractEmbeddedLayer(file *os.File, layerID string, outputPath string) (*os
 	return layerFile, nil
 }
 
+// getAncestry computes an image ancestry, returning an ordered list
+// of dependencies starting from the topmost image to the base.
+// It checks for dependency loops via duplicates detection in the image
```
jonboulle: duplicate detection

lib/internal/backend/file/file.go
```diff
@@ -279,14 +282,24 @@ func extractEmbeddedLayer(file *os.File, layerID string, outputPath string) (*os
...
 	curImgID := imgID
 
 	var err error
 	for curImgID != "" {
+		if deps[curImgID] {
+			return nil, fmt.Errorf("dependency loop detected at image %q", curImgID)
+		}
+		deps[curImgID] = true
 		ancestry = append(ancestry, curImgID)
+		log.Debug(`Getting ancestry for layer "`, curImgID, `".`)
```
jonboulle: IMO nicer to just format the string here: `fmt.Sprintf("... %q ...", curImgID)`

lib/common/common.go
```diff
@@ -60,3 +63,21 @@ func ParseDockerURL(arg string) (*ParsedDockerURL, error) {
...
+func IsValidLayerRefs(name string) error {
+	validName := regexp.MustCompile(`^([A-Fa-f0-9_+.-]+)$`)
+	if ok := validName.MatchString(name); !ok {
+		return fmt.Errorf("invalid layer name %q", name)
+	}
+	return nil
+}
+
+// IsValidLayerId validates a layer ID
```
jonboulle: Naming nit: generally `IsFoo` functions return a `bool`. Instead perhaps `AssertValidLayerId` or `ValidateLayerID`

lucab: My bad, this was just returning a bool at first but then changed to error returning, but I forgot to reflect that in the naming.

lib/common/common.go
lucab: This was supposed to be a library-exposed helper, but I guess we may re-introduce it later as soon as any consumers need it.

lucab:
Further iteration on this: removed unused bits, renamed function and cleaned the debug printing.

lib/common/common.go
```diff
@@ -60,3 +63,12 @@ func ParseDockerURL(arg string) (*ParsedDockerURL, error) {
 	p, err := docker.ParseDockerURL(arg)
 	return (*ParsedDockerURL)(p), err
 }
+
+// ValidateLayerId validates a layer ID
+func ValidateLayerId(id string) error {
+	validId := regexp.MustCompile(`^(\w+:)?([A-Fa-f0-9]+)$`)
```
jonboulle: Sorry, missed this on previous pass: this should a global variable (so failure would be at initialization time, not function invocation time)

lucab: Ack, moved to a global var stanza.

PR #206, opened 2016-10-13
lib/internal: re-use uid value when gid is missing

### LINKED ISSUE #175: GID is set to root when GID is missing
When GID is empty, we use the root GID today: https://github.com/appc/docker2aci/blob/4e051449c0079ba8df59a51c14b7d310de1830b8/lib/internal/internal.go#L466

However this is not great, we should look up the GID from the /etc/passwd file in that image.

Ref:
https://github.com/appc/spec/issues/623
https://github.com/kubernetes/kubernetes/pull/27200

cc @iaguis @alban

lucab:
This PR changes the behavior of uid/gid translation, by re-using
uid value when available as a fallback for a missing gid.

Fixes #175

### DIFF
--- a/lib/internal/internal.go
+++ b/lib/internal/internal.go
@@ -654,9 +654,9 @@ func parseDockerUser(dockerUser string) (string, string) {
 
 	// when only the user is given, the docker spec says that the default and
 	// supplementary groups of the user in /etc/passwd should be applied.
-	// Assume root group for now in this case.
+	// To avoid inspecting image content, we set gid to the same value as uid.
 	if len(dockerUserParts) < 2 {
-		return dockerUserParts[0], "0"
+		return dockerUserParts[0], dockerUserParts[0]
 	}
 
 	return dockerUserParts[0], dockerUserParts[1]


euank:
This still interprets the instruction differently.

Before, we had a security concern (in root group when you wouldn't be under Docker for an image). I think the concern wasn't too big because the root group has few special permissions.

Now we have the chance of images that do not run anymore; e.g. if the `user` given there is a string, not an id, and a group of that name doesn't exist.

Still, better to fail more often than be insecure, so LGTM.

We should open a followup issue to do the right thing eventually though.

euank:
Actually, we should probably check if it's a name or uid. If it's an id, then afaik setting uid never implies setting gid in a docker image.
I think doing so ends up being more correct and we still wouldn't have to deal with reading `passwd`.

lucab:
Thanks for the feedback, but I think I'm not correctly understanding your last comment. user+group are required settings in appc, so we can't set the former but not the latter.

euank:
Oops, I'd forgotten.

LGTM in that case.

PR #207, opened 2016-10-13
lib/internal: do not compare tag when pulling by digest

lucab:
This commit fixes docker2aci pull-by-digest in order to avoid an
improper fallback to "latest" tag and a wrong comparison when no
tags have been specified.

Reference: https://github.com/coreos/rkt/issues/2313#issuecomment-245455406


lucab:
This should have been spotted by `TestFetchingByDigestV22`, so there is probably another bug to be chased in the testsuite.

/cc @dgonyeo

jonboulle:
nice catch, lgtm

PR #209, opened 2016-10-19
Map Docker labels to appc annotations.

grepory:
Given a Dockerfile:

```
FROM scratch
LABEL foo=bar
ADD emptyfile /
```

touch emptyfile

docker build -t grep/test .

docker save -o docker/test.docker grep/test

Exported docker manifest:

```
[
  {
    "Config": "02d28b56668b4dcd4f354b2799f65246378aa23e4f5e62db4d6eb7d8737beee9.json",
    "RepoTags": [
      "grep/test:latest"
    ],
    "Layers": [
      "0e664782202308a03920d0fe133f34e6d348e0cd2cf315c24d71a1c5d8b48f6f/layer.tar"
    ]
  }
]
```

```
cat 02d28b56668b4dcd4f354b2799f65246378aa23e4f5e62db4d6eb7d8737beee9.json | jq .config
{
  "Hostname": "9cfcce2510b1",
  "Domainname": "",
  "User": "",
  "AttachStdin": false,
  "AttachStdout": false,
  "AttachStderr": false,
  "Tty": false,
  "OpenStdin": false,
  "StdinOnce": false,
  "Env": [
    "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
  ],
  "Cmd": null,
  "Image": "sha256:dc7cef81fee136961f5302ddd013a29aaf0a3f04e0dfc2208ed8a53d34bacb8b",
  "Volumes": null,
  "WorkingDir": "",
  "Entrypoint": null,
  "OnBuild": null,
  "Labels": {
    "foo": "bar"
  }
}
```

docker2aci test.docker

tar xvf grep-test-latest.aci

```
cat manifest | jq .annotations
[
  {
    "name": "created",
    "value": "2016-10-06T20:34:17Z"
  },
  {
    "name": "appc.io/docker/repository",
    "value": "grep/test"
  },
  {
    "name": "appc.io/docker/imageid",
    "value": "0e664782202308a03920d0fe133f34e6d348e0cd2cf315c24d71a1c5d8b48f6f"
  },
  {
    "name": "foo",
    "value": "bar"
  }
]
```

### DIFF
--- a/lib/internal/internal.go
+++ b/lib/internal/internal.go
@@ -271,6 +271,10 @@ func GenerateManifest(layerData types.DockerImageData, dockerURL *types.ParsedDo
 
 			genManifest.App = app
 		}
+
+		for k, v := range dockerConfig.Labels {
+			addAnno(k, v)
+		}
 	}
 
 	if layerData.Parent != "" {
--- a/lib/internal/types/docker_types.go
+++ b/lib/internal/types/docker_types.go
@@ -62,6 +62,7 @@ type DockerImageConfig struct {
 	NetworkDisabled bool
 	MacAddress      string
 	OnBuild         []string
+	Labels          map[string]string
 }
 
 // DockerAuthConfigOld represents the deprecated ~/.dockercfg auth


grepory:
I don't understand the nature of this semaphoreci build failure.

jonboulle:
Semaphore was a flake it seems.
This looks okay to me but we probably need to deal with conflicts (e.g. what if a user has set labels that conflict with the ones docker2aci generates?)
@lucab thoughts?

lucab:
I also have the same concern. I think one possible way out of this is:
 1. define the list of our reserved prefixes. At the moment, these seem to be: `created`, `docker-comment`, `author`, `authors`, `appc.io/docker/`.
 2. set out an additional schema-label prefix for ourself (`io.appc.docker2aci`?)
 3. translate all conflicting and dot-free labels to annotations under our prefix (`io.appc.docker2aci.labels`?)
 4. translate all non-conflicting and schema-label scoped entry to annotations, copied over 1:1
 5. bonus: gradually start emitting our own generated annotations under our schema-label prefix

Am I missing some corner case?

lucab:
I've a doubt, so I'm paging @squeed here. I know we have different kinds of annotations and labels now in appc/spec, which one should be used here?

jonboulle:
GH seems to have eaten my comment - reposting:

I forgot to come back to this after the recent appc changes. I think we can potentially just put the docker labels directly into the new userLabels field - the only problem I see is if the CRI/dockershim expects something different (i.e. it mutates Docker labels instead of passing them through), but I doubt it?

jonboulle:
I forgot to come back to this after the recent appc changes. I think we can
potentially just put the docker labels directly into the new userLabels
field - the only problem I see is if the CRI/dockershim expects something
different, but I doubt it?

squeed:
@jonboulle the answer to your question is, as they say, clear as mud: kubernetes/kubernetes#3764

grepory:
so that issue is sort of confusing because it discusses container labels as well as image labels? I think the LABEL directive in Dockerfiles doesn't lead to labels on running containers. Also it looks like k8s adds only runtime labels--since image labels are immutable. 

Can we consider only image labels for the time being? If this userLabels field is strictly for user-provided metadata, it seems like a simple 1to1 mapping appropriate.

lucab:
Superseded by https://github.com/appc/docker2aci/pull/223, closing.

PR #210, opened 2016-10-21
lib/internal: fix vet errors, and error handling

### LINKED ISSUE #194: Go 1.7 vet compliance violation
Currently docker2aci copies the `http#DefaultTransport` struct, and modifies it , see https://github.com/appc/docker2aci/blob/deadf1f9d3a9ca2144efbb65ad223e535f8f9c24/lib/internal/util/util.go#L76.

This causes `go vet` starting from Go 1.7 to complain. In rkt we moved to using an internal client rather than the stdlib defaults.

xref https://github.com/coreos/rkt/pull/3084

s-urbaniak:
This introduces go vet, and fixes error handling in backend/file.

Fixes #194

### DIFF
--- a/lib/internal/backend/file/file.go
+++ b/lib/internal/backend/file/file.go
@@ -364,7 +364,7 @@ func getJsonV22(file *os.File, layerID string) ([]byte, error) {
 func getTarFileBytes(file *os.File, path string) ([]byte, error) {
 	_, err := file.Seek(0, 0)
 	if err != nil {
-		fmt.Errorf("error seeking file: %v", err)
+		return nil, fmt.Errorf("error seeking file: %v", err)
 	}
 
 	var fileBytes []byte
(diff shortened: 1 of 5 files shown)


build.sh
```diff
@@ -30,4 +30,6 @@ if [ "${GOOS}" = "freebsd" ]; then
 fi
 
 echo "Building docker2aci..."
+go vet ./pkg/...
```
jonboulle: can you put this in tests/tests.sh instead?

s-urbaniak:
Unfortunately go 1.6.3 (the version we are building with on Semaphor) has a bug with struct literals, and slices, as reported in https://github.com/golang/go/issues/9171. So I am tempted to disable these `go vet` tests now at all :-/

Another possibility: we bump go 1.7 in Semphore.

build.sh
s-urbaniak: good question, personally I see those `go vet` commands actually part of the "compile" phase independent of any tests, but I can move it over. We have a problem with Go 1.6.3 though in Semaphore (see https://github.com/appc/docker2aci/pull/210#issuecomment-255387525).

s-urbaniak:
Nevermind, I found a way to trick it.

s-urbaniak:
@jonboulle addressed review comments, PTAL

jonboulle:
LGTM, thanks!

Issue #85, opened 2015-09-16
404 when fetching docker image from artifactory hosted docker registry


jaybuff:
It looks like repository.go uses version 1 of the docker registry api [1], but none of the endpoints are prefixed with /v1/.  Here's an example line from repository.go:

    req, err := http.NewRequest("GET", "https://"+path.Join(registry, "repositories", appName, "tags"), nil)

The artifactory docker registry [2] requires that v1 be in the registry url just like the api spec says, but there is no way to specify it to docker2aci.

[1] https://docs.docker.com/v1.6/reference/api/registry_api/
[2] http://www.jfrog.com/confluence/display/RTF/Docker+Repositories

iaguis:
We add the `/v1/` here: https://github.com/appc/docker2aci/blob/8ac8ccffacb8e2040cf94e3af9ca972766353d82/lib/backend/repository/repository.go#L347

jonboulle:
@jaybuff are you seeing docker2aci fail against artifactory?

jaybuff:
Yes, pulling a docker image from artifactory using rkt-0.6.1 results in a 404.  0.6.1 has the line @iaguis references.  I'm investigating why it doesn't work.  Here's the actual command and error:

    $ rkt --insecure-skip-verify fetch docker://registry.example.com/myns/myrepo:2.0
    rkt: fetching image from docker://registry.example.com/myns/myrepo:2.0
    error converting docker image to ACI: error getting ImageID from tag 2.0: HTTP code: 404. URL: https://registry.example.com/repositories/myns/myrepo/tags

jaybuff:
Ah, artifactory 3.9.4 doesn't set the `X-Docker-Endpoints` header in response to the `GET /v1/repositories/myns/myrepo/images`.  I'm not sure if other versions of artifactory set it.

GreatSUN:
Hi all, I am still not able to fetch a Docker image via rkt from Artifactory 4.8.2.
CA Certificate has been added to CoreOS after:

`/var/lib/rkt/rkt fetch docker://<artifactoryhost>:50000/coreos/etcd:v3.0.2 --insecure-options=image
image: remote fetching from URL "docker://<artifactoryhost>:50000/coreos/etcd:v3.0.2"
fetch: Get https://<artifactoryhost>:50000/v2: x509: certificate signed by unknown authority`

After the CA certificate was added, I get:
[...]

PR #214, opened 2016-10-23
backend/repository: assume no v2 on unexpected status

lucab:
This commit changes version detection behavior to align with
Registry v2 specs:

```
If 404 Not Found response status, or other unexpected status,
is returned, the client should proceed with the assumption
that the registry does not implement V2 of the API.
```

Reference: https://docs.docker.com/registry/spec/api/#/api-version-check

Fixes https://github.com/appc/docker2aci/issues/85

### DIFF
--- a/lib/internal/backend/repository/repository.go
+++ b/lib/internal/backend/repository/repository.go
@@ -108,27 +108,26 @@ func (rb *RepositoryBackend) BuildACI(layerIDs []string, dockerURL *types.Parsed
 	}
 }
 
+// checkRegistryStatus determines registry API version compatibility according to spec:
+// https://docs.docker.com/registry/spec/api/#/api-version-check
 func checkRegistryStatus(statusCode int, hdr http.Header, version registryVersion) (bool, error) {
 	switch statusCode {
 	case http.StatusOK, http.StatusUnauthorized:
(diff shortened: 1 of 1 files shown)


lib/internal/backend/repository/repository.go
```diff
@@ -108,27 +108,27 @@ func (rb *RepositoryBackend) BuildACI(layerIDs []string, dockerURL *types.Parsed
...
+			// and clients MAY fallback to v1 if missing, as done here.
 			ok = hdr.Get("Docker-Distribution-API-Version") == "registry/2.0"
 		}
 		return ok, nil
-	case http.StatusNotFound:
+	default:
 		return false, nil
 	}
-
-	return false, fmt.Errorf("unexpected http code: %d", statusCode)
```
s-urbaniak: I think this is a compilation failure, since there is nothing returned.

lucab: Local build+test seems fine (go 1.6.3), but I see over-zealous checkers may get confused. I'll switch to a false-return at the bottom.

lucab:
@s-urbaniak rebased, PTAL.

jonboulle:
LGTM

Issue #216, opened 2016-10-25
lib: allow to specify registry and mediatype options


lucab:
docker2aci (library and standalone binary), should allow the callers to specify a union-set of acceptable registry and mediatype options. This is needed to let consumers specify a certain mediatype to avoid (eg. OCI before stabilization) or certain registry features to enforce (eg. downgrade to registry-v1 or enforce [DCT](https://docs.docker.com/engine/security/trust/content_trust/)).

From the CLI, this may translate into:
- `--registry-options=[registry-v1,registry-v2,registry-v2-dtc]` with an `auto` default covering plain v1+v2 for the moment.
-  `--mediatype-options=[docker-v1,docker-v2.1,docker-v2.2,oci-v1.pre]` with an `auto` default covering docker types for the moment.

For the lib, this will translate into appropriate parameters to all backend types.

lucab:
I did an initial investigation into this, and it seems to require a non-trivial amount of reworking/cleanup both on the internal types and the exposed interface; for these reasons it was deferred post-0.13.

I'm assigning to @s-urbaniak for the design and implementation of the interface/types rework. CLI changes are nice to have but not required at first.

As we are most probably changing the external lib interface, it would be nice to already start to account for:
 - passing a logger interface to the lib: https://github.com/appc/docker2aci/issues
 - different streams for warnings and diagnostic messages: https://github.com/appc/docker2aci/issues/24
 - stream targets set by the caller: https://github.com/appc/docker2aci/issues/61
 - streaming the result of conversion directly, without writing to a file: https://github.com/appc/docker2aci/issues/93

dgonyeo:
I don't like the concept of these flags, docker2aci should be able to auto-determine (and currently can) which mediatypes to use in a given situation. If v2 pulling is supported, it'll do v2 pulling accepting both v2.1 and v2.2 content, and if v2 pulling is unsupported it will automatically fall back to V1 pulling. I'm not sure a user would ever want to selectively disable one of these.

I'm currently making a PR for hiding OCI support behind a flag, but I've got other thoughts on that I'll describe in that PR.

s-urbaniak:
Cross-referencing https://github.com/appc/docker2aci/pull/236#issuecomment-274025517 as it describes the rationale behind this issue.

PR #218, opened 2016-10-26
log: introduce Logger interface

### LINKED ISSUE #136: Allow custom logging in docker2aci lib.
Something to consider before releasing 1.0 in a separate PR - replace the `pkg/log` with a `Logger` field in the `CommonConfig` struct, so users can pass their own logger. The `Logger` fields would be of some interface type like:

``` go
type Logger interface {
    Debug(i ...interface{})
    Info(i ...interface{})
}
```

so the implementer could implement the `Debug` function in the same way as currently `pkg/log` does without resorting to some global boolean variable set by the `InitDebug()` function. Possibly export two implementation like `DefaultLogger` with configurable debugging function and `DevNullLogger`.

s-urbaniak:
This introduces a new Logger interface to allow swappable `debug`, and
`info` loggers for consumers of the library.

These loggers are compatible with the "log" stdlib logger.

This PR also does a refactoring/cleanup in preparation for #216.

Fixes #136


s-urbaniak:
I did this in the plane, so let's see, if I pass the CI.

euank:
LGTM, thanks for the refactoring!

jonboulle:
Needs a rebase

lucab:
Rebased, LGTM, merging. And thanks :)

PR #220, opened 2016-10-27
Correct v2 registry check URL

### LINKED ISSUE #219: Wrong v2 registry check url
Hi there,

It seems that docker2aci has technically been using the wrong v2 registry check URL

Currently it checks: 
`https://registry-1.docker.io/v2`
But the [docs](https://docs.docker.com/registry/spec/api/#api-version-check) specify it should really be:
`https://registry-1.docker.io/v2/`

This has caused us some odd issues with the Sonatype Nexus implementation of docker registry

Here is some example code to explain why the incoming PR is not just adding a `/`
https://play.golang.org/p/fNuI9iVw0S

stuart-warren:
fixes #219
requires a trailing / on the v2 check URL
https://docs.docker.com/registry/spec/api/#api-version-check
some implementations are clearly stricter with the specs
https://play.golang.org/p/fNuI9iVw0S

### DIFF
--- a/lib/internal/backend/repository/repository.go
+++ b/lib/internal/backend/repository/repository.go
@@ -22,7 +22,7 @@ package repository
 import (
 	"fmt"
 	"net/http"
-	"path"
+	"net/url"
 
 	"github.com/appc/docker2aci/lib/common"
 	"github.com/appc/docker2aci/lib/internal/docker"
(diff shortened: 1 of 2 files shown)


jonboulle:
thanks for the fix - LGTM. @lucab ?

lucab:
:+1:

stuart-warren:
Many thanks @jonboulle !
How long until the next release please?
I'd really like to get this into rkt so we can actually use our private Nexus repo rather than pulling big images over a very slow internet link.

jonboulle:
@stuart-warren we typically run on a two-week release cadence - unfortunately we've [_just_](https://github.com/coreos/rkt/pull/3324) cut 1.18.0 today which normally means this'd make it into our 1.19.0 release two weeks from now. I can see an argument for cutting a 1.18.1 with this as a bugfix, but we're pretty constrained on bandwidth right now - if you have any time to put up the patches it might help :-)

lucab:
It looks like this triggered a nefarious side-effect in #221.

PR #222, opened 2016-10-28
lib/internal: allow auto fallback from v2 API to v1

### LINKED ISSUE #221: Need to support docker registries that store v1 and v2 images
Some registries appear to store images that are only accessible via the same registry api version that they were pushed through.

for example:
docker2aci docker://gcr.io/google_containers/kubedns-amd64:1.7 # works!
docker2aci docker://gcr.io/google_containers/skydns:2015-10-13-8c72f8c # 404

Because the code currently assumes that a registry only supports one api version and v2 is checked first, if the registry supports v2 then any v1 images are not accessible.

stuart-warren:
fixes #221
docker2aci would only attempt the v2 api if a registry supported it.
this check for a 404 error from the v2 api and if so attempts to
fallback to v1.

Workflow:
probe v2, try v2, downgrade-on-failure, probe v1, try v1,
hardfail-on-failure

### DIFF
--- a/lib/internal/backend/repository/repository.go
+++ b/lib/internal/backend/repository/repository.go
@@ -94,7 +94,6 @@ func (rb *RepositoryBackend) GetImageInfo(url string) ([]string, *types.ParsedDo
 	var supportsV2, supportsV1, ok bool
 	var URLSchema string
 
-	//probe v2
 	if supportsV2, ok = rb.hostsV2Support[dockerURL.IndexURL]; !ok {
 		var err error
 		URLSchema, supportsV2, err = rb.supportsRegistry(dockerURL.IndexURL, registryV2)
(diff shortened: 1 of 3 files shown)


stuart-warren:
Hmm, I don't really understand the build failure here:
```
fatal: No names found, cannot describe anything.
```

jonboulle:
That's some weird Semaphore flake where I think it doesn't check out the repository properly. I restarted it.

jonboulle:
LGTM. @lucab ?

lucab:
Code LGTM, but I'll give this a spin inside rkt testsuite just to make sure that we don't hit further regressions.

lucab:
rkt testsuite is green too, merging. Thanks a lot @stuart-warren for your patches!

PR #223, opened 2016-11-03
Map Docker labels to appc UserLabels

julia-stripe:
this is a second version of https://github.com/appc/docker2aci/pull/209!

I started out by just converting Docker labels to UserLabels. But then I realized there was a problem! Sometimes, when the Docker image doesn't have a `Cmd` in it, the labels wouldn't get converted.

So I also changed it to not require an image to have a Docker command to convert the other Docker metadata on the image. I can't see why this wouldn't work, but probably there's some reason I don't know yet that this `exec != nil` check was there in the first place.

(also I might have gotten the vendoring wrong -- I haven't used glide before)

### DIFF
--- a/lib/internal/types/docker_types.go
+++ b/lib/internal/types/docker_types.go
@@ -62,6 +62,7 @@ type DockerImageConfig struct {
 	NetworkDisabled bool
 	MacAddress      string
 	OnBuild         []string
+	Labels          map[string]string
 }
 
 // DockerAuthConfigOld represents the deprecated ~/.dockercfg auth
--- a/glide.lock
+++ b/glide.lock
@@ -1,8 +1,8 @@
-hash: 5f32573a0491c3c15d8e6654967c0a9b6798ed25cb9f1ae617f916be1730fb5c
-updated: 2016-10-21T13:40:38.649778864Z
+hash: 1792eab6cf0ce9dc090038e49a5e80de41e5cbb38e06852c41915d1758f27239
+updated: 2016-11-04T06:43:11.742331752-07:00
 imports:
 - name: github.com/appc/spec
-  version: f329150233aa3ac6c83ab5ba02a2b5a8bd03d365
+  version: 2a251f75b2520d7a1c6c10562a1ea4e303af3e67
   subpackages:
   - aci
   - pkg/acirenderer
--- a/glide.yaml
+++ b/glide.yaml
@@ -1,7 +1,7 @@
 package: github.com/appc/docker2aci
 import:
 - package: github.com/appc/spec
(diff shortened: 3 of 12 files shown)


lucab:
Thanks for this PR! Vendoring is correct; there may a bunch of spurious files which are typically removed by glide-vc, have you seen the [helper script](https://github.com/appc/docker2aci/blob/master/scripts/glide-update) to help updating vendored libraries? Regarding the original `exec != nil` check, yes it looks strange and I'm trying to see what's the story behind it. @iaguis or @dgonyeo may have some more knowledge here.

julia-stripe:
Ran the helper script, the extra files should be removed now.

lucab:
Yes, thanks. I'll have a look around to find out what's the deal with the non-nil exec check, and then come back to this PR for a review, sorry for delaying.

lucab:
After a bit of history checking, I think this was introduced at 973d851e02387dcd3e419c42fea6f414a39b1afd and it has been growing out of control from there on. According to [aci spec](https://github.com/appc/spec/blob/master/spec/aci.md#image-manifest-schema) the "exec" field is optional so it is fine to proceed populating other fields if this one is empty. Thanks for bringing up this inconsistency.

lib/internal/types/docker_types.go
```diff
@@ -62,6 +62,7 @@ type DockerImageConfig struct {
 	NetworkDisabled bool
 	MacAddress      string
 	OnBuild         []string
+	Labels          map[string]string
```
lucab: Note to self: this is coming out of https://github.com/docker/docker/blob/master/api/types/container/config.go#L58.

lucab:
Double-checking from my side is finally over, and this LGTM.

I just want to page @euank and @jonboulle here: userLabels are image properties and can be set at build/conversion time (like here). This means that in rkt/k8s we may end up promoting image labels to runtime k8s labels (selectors); I'm not sure if we want this behavior, but my stance is that it should be discussed in rkt/rktlet, as even without docker2aci users may still provide directly built appc/oci images with labels.

jonboulle:
LGTM in itself, and I agree I think we need to solve this in rktlet/CRI in any case.

euank:
LGTM, I don't think that those image labels get raised into k8s at all currently (or in the near future), and I agree that this shouldn't be related so long as we all agree this makes sense just in the context of docker->aci semantics. Which I think we do.

Merge at will (I would if I could)

jonboulle:
@julia-stripe thanks for the contribution!

jonboulle:
@euank invite sent so you can in future

PR #224, opened 2016-11-04
fetch: annotate originally requested name

euank:
Currently, there's no lossless way to actually recover this information.

e.g. if the user fetches: nginx:latest vs library/nginx:latest, the two
outputs are identical.

This provides a way to recover that input.

cc @lucab

### DIFF
--- a/lib/common/common.go
+++ b/lib/common/common.go
@@ -35,6 +35,12 @@ var (
 )
 
 const (
+	// AppcDockerOriginalName is the unmodified name this image was originally
+	// referenced by for fetching, e.g. something like "nginx:tag" or
+	// "quay.io/user/image:latest" This is identical in most cases to
+	// 'registryurl/repository:tag' but may differ for the default Dockerhub
+	// registry or if the tag was inferred as latest.
+	AppcDockerOriginalName  = "appc.io/docker/originalname"
 	AppcDockerRegistryURL   = "appc.io/docker/registryurl"
 	AppcDockerRepository    = "appc.io/docker/repository"
 	AppcDockerTag           = "appc.io/docker/tag"
@@ -48,10 +54,11 @@ const defaultTag = "latest"
 
 // ParsedDockerURL represents a parsed Docker URL.
 type ParsedDockerURL struct {
-	IndexURL  string
-	ImageName string
-	Tag       string
-	Digest    string
+	OriginalName string
+	IndexURL     string
+	ImageName    string
+	Tag          string
+	Digest       string
 }
 
 type ErrSeveralImages struct {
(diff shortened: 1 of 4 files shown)


lib/tests/v22_test.go
```diff
@@ -49,80 +49,86 @@ var dockerConfig = typesV2.ImageConfig{
...
-			Name:  *types.MustACIdentifier("version"),
-			Value: "v0.1.0",
+func expectedManifest(registryUrl, imageName string) schema.ImageManifest {
+	return schema.ImageManifest{
+		ACKind:    types.ACKind("ImageManifest"),
+		ACVersion: schema.AppContainerVersion,
+		Name:      *types.MustACIdentifier("variant"),
+		Labels: []types.Label{
+			types.Label{
+				*types.MustACIdentifier("arch"),
```
jonboulle: tediously, you'll need to key these to shut up govet

lib/common/common.go
```diff
@@ -35,6 +35,12 @@ var (
 )
 
 const (
+	// AppcDockerImageName is the unmodified name this image was original
```
jonboulle: originally

lib/tests/v22_test.go
euank: Whoops, sloppy rebase I think.

jonboulle:
LGTM though I don't get the tests thing.

lib/common/common.go
```diff
@@ -35,6 +35,12 @@ var (
 )
 
 const (
+	// AppcDockerImageName is the unmodified name this image was originally
+	// referenced by for fetching, e.g. something like "nginx:tag" or
+	// "quay.io/user/image:latest" This is identical in most cases to
+	// 'registryurl/repository:tag' but will differ for the Dockerhub registry or
+	// if the tag was inferred as latest.
+	AppcDockerImageName     = "appc.io/docker/imagename"
```
lucab: I find this a bit confusing, as we already use `ImageName` for something else as a field name. Can we mirror more closely the `RawName` field name here?

lib/tests/v22_test.go
```diff
@@ -217,53 +226,44 @@ func manifestEqual(manifest, expected *schema.ImageManifest) error {
...
+	if len(manifest.Annotations) != len(expected.Annotations) {
+		return fmt.Errorf("annotations not equal: %v != %v", manifest.Annotations, expected.Annotations)
 	}
-	exp, _ := expected.GetAnnotation(name)
-	if got != exp {
-		return fmt.Errorf("expected annotation %q, got %q", exp, got)
+	for _, ann := range manifest.Annotations {
+		ea, ok := expected.Annotations.Get(ann.Name.String())
+		if ea == "variant" {
+			// Unknown value, always okay
```
euank: Before, it only asserted on a subset of values (which didn't include `appc.io/docker/imageid`). However, the test image generated doesn't appear to be idempotent so this varies per test.

The expected value is set to this arbitrary value to skip asserting on it.

lib/common/common.go
euank: `rawname` seems less clear, but I agree it's confusing. What about `pulled-name`? Something else?

lucab: I'm open to whatever renaming you think fits here, as long as it can be used consistently in consumer code and we don't step into existing names. I'm non-native, so some of those may be a bit off, but here are some possibilities: `originalname`, `externalname`, `intialname`, `nativename` or such.

euank: I picked `appc.io/docker/originalname`. Thanks for the suggestions!

PR #225, opened 2016-11-04
build: use git describe fallback

lucab:
Pass `--always` to `git describe`, falling back to commit revision
in case tags were not fetched.
This workarounds some non-deterministic race/failure on semaphore:
```
fatal: No names found, cannot describe anything.
```


jonboulle:
lgtm

PR #226, opened 2016-11-07
Oci rc2

vishvananda:
This updates the oci types to rc2. It also does an update of the glide dependencies. The secondary updates are in a separate commit to make it easier to review.


lucab:
Note to self: the large vendor diff is due to unpinned dependencies in appc/spec. Most of those are cleanup for files that became unused.

lucab:
LGTM, thanks for the update!

jonboulle:
<3 thanks!

PR #227, opened 2016-11-08
scripts: update glide-update vendor script

s-urbaniak:
The current glide update flags are deprecated, and the current settings
are default behavior in glide.

This fixes it.


lucab:
LGTM

PR #229, opened 2016-12-02
Image from file improvements; guesses at "originalname" and fixes for "--image"

euank:
The use case is rktlet wanting the `originalname` value when using `docker2aci` + `rkt fetch`.

Originally I figured I could use `--image=...` to have `originalname` end up there rather than guessing at it, but `--image` was broken for any images that need special handling (namely all library/ dockerhub images), and after thinking about it a bit more I think taking a guess at the originalname is okay here.

I left a fairly detailed commit comment for the most complicated of the changes, and the commits are broken up into nice bite-sized diffs.

### DIFF
--- a/lib/internal/backend/file/file.go
+++ b/lib/internal/backend/file/file.go
@@ -54,10 +54,15 @@ func NewFileBackend(file *os.File, debug, info log.Logger) *FileBackend {
 }
 
 func (lb *FileBackend) GetImageInfo(dockerURL string) ([]string, *common.ParsedDockerURL, error) {
-	parsedDockerURL, err := common.ParseDockerURL(dockerURL)
-	if err != nil {
-		// a missing Docker URL could mean that the file only contains one
-		// image, so we ignore the error here, we'll handle it in getImageID
+	// a missing Docker URL could mean that the file only contains one
+	// image so it's okay for dockerURL to be blank
+	var parsedDockerURL *common.ParsedDockerURL
+	if dockerURL != "" {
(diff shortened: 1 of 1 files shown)


lib/internal/backend/file/file.go
```diff
@@ -260,7 +276,7 @@ func getImageID(file *os.File, dockerURL *common.ParsedDockerURL, name string, d
 
 			if dockerURL == nil {
 				dockerURL = &common.ParsedDockerURL{
-					OriginalName: "",
+					OriginalName: appName + ":" + tag,
```
euank: Does this break for images saved from `foo@sha256:$hash`? Can you even save those without tagging them to some string?

lucab: I think that is equivalent to export an image by its raw sha, in which case you don't get a repositories file with name and tag.

lucab:
The `--image` fix is fine. The rest of this is probably not super-nice and may have some further corner-cases, but looks reasonable overall. LGTM, feel free to self-merge if satisfied or add further comments.

lib/internal/backend/file/file.go
euank: If you save an image then you get a manifest.json with a null RepoTags, and docker2aci can't do anything with it currently (errors out, even if you use the `-image` flag).

euank:
After sleeping on it a bit more, I removed the `originalname` guessing. I'm happy with this now, other than the lack of testing ☹️

PR #233, opened 2017-01-11
Bump appc/spec library version to 0.8.9

ybubnov:
This patch updates the version of "github.com/appc/spec" library to
0.8.9 in order to use the introduced function "ToAppcOSArch" for
conversion of Docker image configuration to compatible AppC labels.


ybubnov:
Update of the dependencies required in order to fix https://github.com/coreos/rkt/issues/3526

lucab:
Great, thanks a lot! Do you also plan to wire-in the label conversion? We can tag a release quickly after that so that it can reach to rkt in a short timeframe.

ybubnov:
Yes, I will try to create a required patch in a day or two.

Issue #195, opened 2016-08-25
lib: wrong arch label values on non-amd64 architectures


onlyjob:
_docker2aci_ FTBFS on multiple architectures in Debian, all due to "bad arch" in `TestFetchingMultipleLayersV22`:

i686 (x86_32):

```
--- FAIL: TestFetchingMultipleLayersV22 (0.00s)
    server.go:16: path requested: /v2
    server.go:16: path requested: /v2/docker2aci/dockerv22test/manifests/v0.1.0
    server.go:16: path requested: /v2/docker2aci/dockerv22test/blobs/sha256:e1d3a8cbfefdfa9cae6933f464edd653140cfdef3188e71a23ae5bea6e3acf2c
    server.go:16: path requested: /v2/docker2aci/dockerv22test/blobs/sha256:23929dd118cf976d950e8e89117b5a62908550140803df4feccc24fc26d173b7
    server.go:16: path requested: /v2/docker2aci/dockerv22test/blobs/sha256:44a6f6185deced11e94e69e55352f1fbc5346e5c9bf95f3017ca37d8b064db9c
    v22_test.go:326: error generating ACI: error writing manifest: json: error calling MarshalJSON for type schema.ImageManifest: json: error calling MarshalJSON for type types.Labels: bad arch "386" for linux (must be one of: [amd64 i386 aarch64 aarch64_be armv6l armv7l armv7b])
FAIL
```

arm64:

```
--- FAIL: TestFetchingMultipleLayersV22 (0.02s)
    server.go:16: path requested: /v2
    server.go:16: path requested: /v2/docker2aci/dockerv22test/manifests/v0.1.0
    server.go:16: path requested: /v2/docker2aci/dockerv22test/blobs/sha256:d43552b26516a1c2ce5e54828b7181667ff0eb6e43f78c9ecdd1176c4608a748
    server.go:16: path requested: /v2/docker2aci/dockerv22test/blobs/sha256:ac45c31629f1ded51db473f6669a9a602075ccb397fe166900a451d36a00e2bc
    server.go:16: path requested: /v2/docker2aci/dockerv22test/blobs/sha256:7e5fefe0ff9c239c096c86c7406631a3f43aee418d15508be279908987af272a
[...]

PR #234, opened 2017-01-12
Translate "os" and "arch" labels of image manifest

ybubnov:
This patch adds additional logic of operation system and architecture
labels translation during the conversion of Docker images to ACIs.

It also ships a set of unit tests in order to validate the changes.

References https://github.com/appc/docker2aci/issues/195

### DIFF
--- a/lib/internal/internal.go
+++ b/lib/internal/internal.go
@@ -166,6 +166,42 @@ func generateEPCmdAnnotation(dockerEP, dockerCmd []string) (string, string, erro
 	return entrypointAnnotation, cmdAnnotation, nil
 }
 
+// setLabel sets the label entries associated with non-empty key
+// to the single non-empty value. It replaces any existing values
+// associated with key.
+func setLabel(labels map[appctypes.ACIdentifier]string, key, val string) {
+	if key != "" && val != "" {
+		labels[*appctypes.MustACIdentifier(key)] = val
+	}
+}
+
+// setOSArch translates the given OS and architecture strings into
+// the compatible with application container specification and sets
+// the respective label entries.
+//
+// Returns an error if label translation fails.
+func setOSArch(labels map[appctypes.ACIdentifier]string, os, arch string) error {
+	// Translate arch tuple into the appc arch tuple.
+	appcOS, appcArch, err := appctypes.ToAppcOSArch(os, arch, "")
+	if err != nil {
+		return err
+	}
+
+	// Set translated labels.
+	setLabel(labels, "os", appcOS)
+	setLabel(labels, "arch", appcArch)
+	return nil
+}
+
+// setAnnotation sets the annotation entries associated with non-empty
+// key to the single non-empty value. It replaces any existing values
+// associated with key.
+func setAnnotation(annotations *appctypes.Annotations, key, val string) {
+	if key != "" && val != "" {
+		annotations.Set(*appctypes.MustACIdentifier(key), val)
+	}
+}
+
 // GenerateManifest converts the docker manifest format to an appc
 // ImageManifest.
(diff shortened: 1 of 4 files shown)


ybubnov:
This patch is a next step to fix https://github.com/coreos/rkt/issues/3526. Previously, ```appc/spec``` library have been updated to ```0.8.9``` version: https://github.com/appc/docker2aci/pull/233

lib/internal/internal.go
```diff
@@ -195,42 +231,23 @@ func GenerateManifest(layerData types.DockerImageData, dockerURL *common.ParsedD
...
-	}
+	setLabel(labels, "layer", layerData.ID)
+	setLabel(labels, "version", dockerURL.Tag)
 
-	addLabel("layer", layerData.ID)
-	addLabel("version", dockerURL.Tag)
-	addLabel("os", layerData.OS)
-	addParentLabel("os", layerData.OS)
-	addLabel("arch", layerData.Architecture)
-	addParentLabel("arch", layerData.OS)
```
ybubnov: Not sure, if ```arch``` label supposed to be mapped to ```OS```, or just a mistake.

lucab: I think this was by mistake. Pinging @dgonyeo and @jonboulle just to double check if there is anything tricky I'm missing here.

 Moreover, it looks like `parentLabels` is never actually used after the assignment, am I reading this right?

lib/internal/internal_test.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func TestOSArch(t *testing.T) {
+	tests := []struct {
+		srcOS, srcArch string
+		dstOS, dstArch string
+		err            bool
+	}{
+		{"", "", "", "", false},
+		{"TempleOS", "ia64", "", "", false},
```
lucab: :wink:

lib/internal/internal.go
dgonyeo: Yes, I believe that was a mistake. Also if memory serves, `parentLabels` is just set in case a user wishes to inspect the ACI later.

lucab: @dgonyeo but from here it doesn't seem to be actually persisted into the ACI manifest, uh?

lucab:
I think `lib/tests/v22_test.go` may also need some adjustment in order to avoid label mismatch.

lucab:
@glevand @trusch care to give this a spin on aarch64? It should spit out an image with proper appc os/arch tuple. I'm not sure about the testsuite though.

glevand:
Seems OK.

before:
```
$ ./bin/docker2aci -debug docker://aarch64/busybox
Getting image info...
Downloading sha256:cd56a5768f0 [=============================] 1.77 MB / 1.77 MB
Generating layer ACI...
Error: conversion error: error generating ACI: bad arch "arm64" for linux (must be one of: [amd64 i386 aarch64 aarch64_be armv6l armv7l armv7b ppc64 ppc64le s390x])
```
after:
```
$ ./bin/docker2aci -debug docker://aarch64/busybox
Getting image info...
Downloading sha256:cd56a5768f0 [=============================] 1.77 MB / 1.77 MB
Generating layer ACI...
Squashing layers...
Rendering ACI...
Writing squashed ACI...
Validating squashed ACI...
ACI squashed!

Generated ACI(s):
aarch64-busybox-latest.aci
```

lucab:
@glevand thanks. Testsuite as well?

ybubnov:
> I think lib/tests/v22_test.go may also need some adjustment in order to avoid label mismatch.

@lucab, you mean, add a test e.g. with a successful fetch of the image made for non-amd64 arch?

lucab:
@ybubnov yep, basically applying label translation to tests as well instead of harcoding amd64 in there.

ybubnov:
Ack, I will do this.

ybubnov:
Modified the tests in ```v22_test.go``` in order to verify all supported architectures.

PR #236, opened 2017-01-19
Move OCI support behind experimental flag

dgonyeo:
This commit adds a boolean flag to both the library and cli interfaces
to enable OCI support, that defaults to off. OCI support is
enabled/disabled by altering the Accept HTTP headers on requests
(specifically the OCI media types are added/omitted appropriately).

This is necessary for https://github.com/coreos/rkt/issues/3134 to be implemented in rkt, and is related to https://github.com/appc/docker2aci/issues/216.

### DIFF
--- a/lib/docker2aci.go
+++ b/lib/docker2aci.go
@@ -65,9 +65,10 @@ func (c *CommonConfig) initLogger() {
 // converting Docker images.
 type RemoteConfig struct {
 	CommonConfig
-	Username string                // username to use if the image to convert needs authentication
-	Password string                // password to use if the image to convert needs authentication
-	Insecure common.InsecureConfig // Insecure options
+	Username   string                // username to use if the image to convert needs authentication
+	Password   string                // password to use if the image to convert needs authentication
+	Insecure   common.InsecureConfig // Insecure options
+	OCISupport bool                  // Experimental flag to enable OCI support
 }
 
 // FileConfig represents the saved file specific configuration for converting
@@ -94,6 +95,7 @@ func ConvertRemoteRepo(dockerURL string, config RemoteConfig) ([]string, error) 
 			config.Password,
 			config.Insecure,
 			config.Debug,
+			config.OCISupport,
 		),
 		dockerURL: dockerURL,
 		config:    config.CommonConfig,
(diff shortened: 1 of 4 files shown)


dgonyeo:
I'm making this PR for the sake of the rkt issue, but I think I'd prefer to leave things the way they currently are. When doing v2.2 pulling, docker2aci just signifies to the registry that it can accept OCI images in addition to docker images, so leaving in OCI support wouldn't interfere with docker2aci's ability to pull docker images in any way, and when registries start supporting OCI images it'll just work without the users needing to dig for some flag or us needing to flip a switch in some update.

lucab:
I may agree on your concerns regarding docker2aci binary UX, but I don't agree from the library side.

OCI auto-detection is just a small part of this. The larger picture is that we are aware that: 
 * there are old/broken registry implementations in the wild, like [artifactory](https://github.com/appc/docker2aci/issues/85) and some more reported privately. 
 * some registries do automatic v1<->v2 mapping, while some other don't. Our heuristic already broke in the past, like in https://github.com/appc/docker2aci/pull/220.
 * registries are proceeding with deprecation plans, eg. [GCR dropping v1](https://cloud.google.com/container-registry/docs/support/deprecation-notices). Clients should be able to check for this before the cut-off date by disabling v1 explicitely.
 * once we grow support for signed images (DTC or whatever OCI will have), auto-downgrading won't always be the correct approach.
 * Even when the "experimental" flag is removed from OCI, I expect to still encounter broken implementations in the wild and an evolving protocol spec.
 * same reasoning applies for local tarballs. Library clients who already know the source type, should be able to enforce that to avoid surprises (security-wise and more).

I agree that we should keep using our heuristic by default, but library users should have some way of controlling what's going on behind the curtains and select/drop specific modes. Hence the request in #216. That will also make it easier for us to move forward in adding OCI and dropping v1 at some point, without breaking things in a non-overrideable way.

dgonyeo:
Ah ok, I wasn't aware of all the registry issues. I'll go implement https://github.com/appc/docker2aci/issues/216, but if I'm doing that this PR seems redundant (or would at least be implemented differently), so I'm going to close this.

PR #237, opened 2017-01-20
Annotate manifest hash

dgonyeo:
There are two commits in here. In order they:
1. Add an annotation to produced ACIs with a hash of the original docker manifest (or sometimes the app ID depending on the version of the HTTP API used)
2. tweaks the library API so that callers can provide a list of manifest hashes they already have

These two changes should allow for rkt to provide a list of docker manifest hashes (obtained by ACI manifest annotations) for images it already has but wants docker2aci to check for an update for. If docker2aci pulls the manifest for an image and it's in the list of images already owned by the caller, it doesn't have to download the image.

This is part of the work for https://github.com/coreos/rkt/issues/2937

### DIFF
--- a/lib/internal/backend/file/file.go
+++ b/lib/internal/backend/file/file.go
@@ -53,7 +53,13 @@ func NewFileBackend(file *os.File, debug, info log.Logger) *FileBackend {
 	}
 }
 
-func (lb *FileBackend) GetImageInfo(dockerURL string) ([]string, *common.ParsedDockerURL, error) {
+// GetImageInfo, given the url for a docker image, will return the
+// following:
+// - []string: an ordered list of all layer hashes
+// - string: a unique identifier for this image, like a hash of the manifest
+// - *common.ParsedDockerURL: a parsed docker URL
+// - error: an error if one occurred
+func (lb *FileBackend) GetImageInfo(dockerURL string) ([]string, string, *common.ParsedDockerURL, error) {
 	// a missing Docker URL could mean that the file only contains one
 	// image so it's okay for dockerURL to be blank
 	var parsedDockerURL *common.ParsedDockerURL
@@ -61,7 +67,7 @@ func (lb *FileBackend) GetImageInfo(dockerURL string) ([]string, *common.ParsedD
 		var err error
 		parsedDockerURL, err = common.ParseDockerURL(dockerURL)
 		if err != nil {
-			return nil, nil, fmt.Errorf("image provided couldnot be parsed: %v", err)
+			return nil, "", nil, fmt.Errorf("image provided couldnot be parsed: %v", err)
 		}
 	}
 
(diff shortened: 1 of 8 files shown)


dgonyeo:
Some test is unhappy, I'll fix it on Monday

lib/internal/backend/file/file.go
```diff
@@ -53,15 +53,15 @@ func NewFileBackend(file *os.File, debug, info log.Logger) *FileBackend {
 	}
 }
 
-func (lb *FileBackend) GetImageInfo(dockerURL string) ([]string, *common.ParsedDockerURL, error) {
+func (lb *FileBackend) GetImageInfo(dockerURL string) ([]string, string, *common.ParsedDockerURL, error) {
```
s-urbaniak: this return signature is getting messy, it is not obvious what `[]string, string` is. Can we document or name the return values?

lib/internal/internal.go
```diff
@@ -53,13 +53,13 @@ import (
 // BuildACI takes a Docker layer, converts it to ACI and returns its output
 // path and its converted ImageManifest.
 type Docker2ACIBackend interface {
-	GetImageInfo(dockerUrl string) ([]string, *common.ParsedDockerURL, error)
-	BuildACI(layerIDs []string, dockerURL *common.ParsedDockerURL, outputDir string, tmpBaseDir string, compression common.Compression) ([]string, []*schema.ImageManifest, error)
+	GetImageInfo(dockerUrl string) ([]string, string, *common.ParsedDockerURL, error)
```
s-urbaniak: same comment applies here as above. can we document what `[]string, string` are?

lib/internal/internal.go
```diff
@@ -346,7 +347,7 @@ func GenerateEmptyManifest(name string) (*schema.ImageManifest, error) {
 	}, nil
 }
 
-func GenerateManifestV22(name string, dockerURL *common.ParsedDockerURL, config *typesV2.ImageConfig, imageDigest string, lowerLayers []*schema.ImageManifest, debug log.Logger) (*schema.ImageManifest, error) {
+func GenerateManifestV22(name, manhash string, dockerURL *common.ParsedDockerURL, config *typesV2.ImageConfig, imageDigest string, lowerLayers []*schema.ImageManifest, debug log.Logger) (*schema.ImageManifest, error) {
```
s-urbaniak: woah, this is getting immense ;-) at least we could make this a newline separated arglist and squash `imageDigest` with `name, manhash`. Also some documentation would welp. I am becoming concerned about the size of this arglist.

dgonyeo:
Fixed the test, and added some comments to hopefully address the nits.

euank:
I'd prefer using the upstream 'sha256' hash if we can since it'll make things a bit nicer for the user to read.

It also has one small benefit! If the user does `rkt fetch docker://foo@sha256:....` and we annotated that, we can know that we don't need to do any network traffic at all.

dgonyeo:
I agree with @euank, so please don't merge this until I can update that logic. I've got the rkt PR based on this ready though, so as soon as this goes in the rest of https://github.com/coreos/rkt/issues/2937 will hopefully go pretty quickly.

dgonyeo:
This now should have the behavior I want, but https://github.com/appc/docker2aci/pull/238 should be merged first so I don't have to include that commit in this PR.

dgonyeo:
Since @lucab helped me with vendoring, I just removed the vendoring commits and re-pushed this.

dgonyeo:
Bump on this, I have a PR for rkt once this is done :)

PR #238, opened 2017-01-24
vendor: added missing `golang.org/x/net`

dgonyeo:
I was trying to add a package via glide for a feature, and glide kept breaking due to missing packages from `golang.org/x/net`. I've added that package to glide.yaml and then ran `glide up -v` and `glide vc --only-code --no-tests`


lucab:
I think something went horribly wrong in here, but I'm not sure what. It seems to be pulling in weird things like aws-sdk, jose and grpc. I re-did the original go-digest add in #238 (via `scripts/glide-update`) and seems much saner. Can you try to build on top of that? I think the root of most issues around glide here is that appc imports kubernetes.

dgonyeo:
Closed in favor of @lucab's PR

PR #239, opened 2017-01-24
Allow selective disabling of registries and media types

dgonyeo:
This commit tweaks the library's API to allow consumers to selectively
disable registry support (v1 or v2) and what mediatypes docker2aci
claims to be able to accept during an image pull (v2.1, v2.2, and oci).
By default all options are enabled, and this is still the case when
using docker2aci via the command line.

Fixes https://github.com/appc/docker2aci/issues/216.

### DIFF
--- a/lib/common/common.go
+++ b/lib/common/common.go
@@ -21,6 +21,8 @@ import (
 
 	"github.com/appc/docker2aci/lib/internal/docker"
 	"github.com/docker/distribution/reference"
+
+	spec "github.com/opencontainers/image-spec/specs-go/v1"
 )
 
 type Compression int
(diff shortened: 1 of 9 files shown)


dgonyeo:
And for the sake of cross linking, this PR will allow https://github.com/coreos/rkt/issues/3134 to be implemented.

dgonyeo:
Oh and for more cross linking, it's a continuation of https://github.com/appc/docker2aci/pull/236, in which there's more explanation on the rationale of these changes.

main.go
```diff
@@ -37,6 +37,7 @@ var (
 	flagInsecureAllowHTTP  bool
 	flagCompression        string
 	flagVersion            bool
+	flagOCISupport         bool
```
lucab: I think this is not consumed anymore in the PR. Either you forgot to scratch this out, or there are some changes missing from this PR.

main.go
```diff
@@ -47,6 +48,7 @@ func init() {
 	flag.BoolVar(&flagInsecureAllowHTTP, "insecure-allow-http", false, "Uses unencrypted connections when fetching images")
 	flag.StringVar(&flagCompression, "compression", "gzip", "Type of compression to use; allowed values: gzip, none")
 	flag.BoolVar(&flagVersion, "version", false, "Print version")
+	flag.BoolVar(&flagOCISupport, "experimental-oci", false, "Enable experimental OCI support")
```
lucab: I'm perfectly fine about not surfacing the internal complexity of the library regarding mediatype and registry, and just having a single switch for OCI if you prefer.

As I think we are already doing this silently now and it's going to be the default mode at some point, perhaps it should instead be a switch which limits to dockerv1/v2 only?

lucab:
Thinking aloud: I think the library consumer may be interested in knowing how an image has been pulled (or it may not? @s-urbaniak @euank).

If that is a useful information, we can either return the type back to the caller (but then it needs to be persisted somehow), or directly encode it as an image annotation.

dgonyeo:
If we're going to inform docker2aci consumers how an image was pulled I'd prefer to encode it as an image annotation. We already encode a lot of information like this, and it means you can go back and re-inspect the value at any point.

Still, I'd prefer to get some confirmation this is useful before I do the work :)

main.go
dgonyeo: Whoops, forgot to remove this. This is leftover from when the PR was to add a flag to disable OCI support. Removing now.

main.go
dgonyeo: Same as above, I didn't intend to include this.

> just having a single switch for OCI if you prefer

Wasn't that what this PR was: https://github.com/appc/docker2aci/pull/236 ?

I intended to not change the cli at all, and library users can easily ignore the added complexity if they want to. The zero value of the fields added to the struct used by library consumers result in the same behavior as before (as in, they don't set the fields and the same thing happens as before this change). If consumers wish to disable support for one of the registry HTTP APIs or specific media types in the latest API version, then they simply provide a list of what they _do_ want docker2aci to support.

s-urbaniak:
The information how an image has been pulled can be interesting as an intermediate step in the transition to native image support in rkt, but currently I am not aware of its usefulness. Since it's doable easily I suggest we introduce such a feature when explicitely needed.

lucab:
I agree that is both easy to add later and not required now, so let's just skip for the moment.

lib/common/common.go
```diff
@@ -112,3 +114,143 @@ func ValidateLayerId(id string) error {
...
+	MediaTypeOptionDockerV21 = iota
+	MediaTypeOptionDockerV22
+	MediaTypeOptionOCIV1Pre
+)
+
+// MediaTypeSet represents a set of media types which docker2aci is to use when
+// fetchimg images. As an example if a MediaTypeSet is equal to
+// {MediaTypeOptionDockerV22, MediaTypeOptionOCIV1Pre}, then when an image pull
+// is made V2.1 images will not be fetched. This doesn't apply to V1 pulls. As
+// an edge cast if a MedaTypeSet contains nothing, that means that _every_ type
```
s-urbaniak: @lucab yes I think it's fine to have the zero value (nil) or empty list as a valid value. I don't think we need to have an explicit entry.

lib/common/common.go
```diff
@@ -112,3 +114,143 @@ func ValidateLayerId(id string) error {
...
+
+	MediaTypeDockerV22Manifest     = "application/vnd.docker.distribution.manifest.v2+json"
+	MediaTypeDockerV22ManifestList = "application/vnd.docker.distribution.manifest.list.v2+json"
+	MediaTypeDockerV22Config       = "application/vnd.docker.container.image.v1+json"
+	MediaTypeDockerV22RootFS       = "application/vnd.docker.image.rootfs.diff.tar.gzip"
+
+	MediaTypeOCIManifest     = spec.MediaTypeImageManifest
+	MediaTypeOCIManifestList = spec.MediaTypeImageManifestList
+	MediaTypeOCIConfig       = spec.MediaTypeImageConfig
+	MediaTypeOCILayer        = spec.MediaTypeImageLayer
```
lucab: To match the name of the imported package, I think thouse should have a "V1" somewhere in them (eg. `MediaTypeOCIV1Layer`)

lib/common/common.go
```diff
@@ -112,3 +114,143 @@ func ValidateLayerId(id string) error {
...
+// spec.
+type MediaTypeOption int
+
+const (
+	MediaTypeOptionDockerV21 = iota
+	MediaTypeOptionDockerV22
+	MediaTypeOptionOCIV1Pre
+)
+
+// MediaTypeSet represents a set of media types which docker2aci is to use when
```
dgonyeo: You're right in that nothing particularly bad will happen if the set invariant is violated, but it would probably be best for me to mention it regardless.

dgonyeo:
All nits addressed

dgonyeo:
@lucab bump on this

lucab:
I think that except for a rebase and some tests adjustments, this is ready to go.

dgonyeo:
Rebased and fixed tests

Issue #245, opened 2017-04-23
Manifest differ with same image


gfyrag:
When converting a remote image, i ended up with two differents manifest (Not all the times).
There is just labels which are not write with the same order ("os" and "arch" in my case).
The results is two aci files with differents signatures.

Using rocket, i discover this when fetching "latest" image multiple times.
This produce multiple images with the same name in the store but with sha256 hash differing.

I don't know if this can be enforced or if it can be considered as a bug...

lucab:
golang maps are unordered collections. JSON as well does not guarantee any ordering for dict entries. If we need to maintain the hash-identity property on converted artifacts, we should switch to some canonical JSON form.

euank:
Over in OCI land, the image spec has a "SHOULD use canonical json" [here](https://github.com/opencontainers/image-spec/blob/319fb0a1b9831ec33978f905c1a17e7e4084438e/considerations.md#json).

I believe docker also uses that, so we can both use their library and act more like a typical docker client by doing so.

cc @dgonyeo

lucab:
@euank thanks for the reference. Just FYI, I got [dubious datapoints](https://github.com/docker/distribution/blob/f180e9a9349040229a5a72b5867e098600cf9b55/docs/spec/json.md#go) when looking at docker before.

But [canonical/json](https://godoc.org/github.com/docker/go/canonical/json) seems fine, well contained and easy to drop-in. @gfyrag would you like to try tackling your issue by switching to the `Canonical` encoder provided by the above package?

gfyrag:
@lucab Ok, i'm not familiar with the code but i'll give it a try this week end.

lucab:
@gfyrag great! No hurry, take time to learn and make the process enjoyable, and feel free to ask.

Just for reference, [this](https://github.com/appc/docker2aci/pull/242) is how we typically add dependencies (i.e. via an helper [script](https://github.com/appc/docker2aci/blob/master/scripts/glide-update)).

And the package you are looking for should be around [here](https://github.com/appc/docker2aci/blob/1f3b0309824483efc8b0419e26460c8849fd5b9b/lib/internal/internal.go#L722) (from a quick skim, but please double-check before starting).

gfyrag:
@lucab Hello. Here the PR #247 (After a very long delay...)

I integrated canonical/json package. I just change the code where the manifest is write.
Maybe change other parts ?

[...]

PR #247, opened 2017-06-12
Change library used to encode manifest.

gfyrag:
Manifest was encoded using official encoding/json which does not canonical encode data.
Use library from github.com/docker/go/canonical/json instead.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,7 @@ tests/*.docker
 
 *.swp
 *.aci
+
+# Intellij files
+*.iml
+.idea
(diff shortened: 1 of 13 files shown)


onlyjob:
I'm concerned about use of [github.com/docker/go](https://github.com/docker/go). The library is uncommon; it is used (almost?) exclusively by Docker's [Notary](https://github.com/theupdateframework/notary) and appears to be poorly maintained or dormant (the only release in 2015, last commit 2+ years ago)...

Perhaps it would be best to avoid this library.

lucab:
For reference, this PR comes from the discussion at https://github.com/appc/docker2aci/issues/245.

PR #248, opened 2017-06-12
Avoid panicing on scratch images

euank:
Docker allows creating empty images (e.g. with FROM scratch followed by
a no-op like MAINTAINER or LABEL).

Without this patch, docker2aci will panic when it encounters an image of
this sort.

These images are more-or-less useless. They're also difficult to work
with. As an example, `docker save` bails on them with an `unimplemented`
error.

In line with that behavior, I think bailing in docker2aci makes sense
too rather than dealing with more robust error handling or creating an
empty ACI.

### DIFF
--- a/lib/docker2aci.go
+++ b/lib/docker2aci.go
@@ -150,6 +150,9 @@ func (c *converter) convert() ([]string, error) {
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


euank:
`sudo rkt fetch --insecure-options=image docker://euank/scratch` if you want to see the panic 😄

PR #250, opened 2017-09-18
lib/common: update `ParseDockerURL`

schu:
When testing rktlet with k8s `v1.8.0-beta.1`, `PullImage` currently
fails because the kubelet sends a full image name (e.g.
`docker.io/library/busybox`) instead of an abbreviated one (e.g.
`busybox`). This is probably due to a regression in k8s; see also
kubernetes/kubernetes#52110

By updating `github.com/docker/distribution/reference`, we can use new
helpers to parse and harmonize names. With that, docker2aci now also can
handle URLs with the default index domain (`docker.io`) in it, for
example `docker://docker.io/library/busybox`. Before, that resulted in
`Error: conversion error: registry doesn't support API v2 nor v1`, as
`docker.io` doesn't point to a registry API.

```
docker://busybox
docker://library/busybox
docker://docker.io/library/busybox # didn't work before
```

This works because `docker.io` as the default index domain gets stripped
from a familiarized name and thus docker2aci uses registry-1.docker.io
instead:

https://github.com/docker/distribution/blob/5db89f0ca68677abc5eefce8f2a0a772c98ba52d/reference/normalize.go#L90-L91


lucab:
Can you please add some basic unit testing to this? Also, does this introduce a delta in the generated image manifest of the final ACI, compare to previous behavior?

iaguis:
yep, unit testing would be good :)

lucab:
Just for the sake of clarity, the image-manifest delta I'm worried about is for the following cases:
 * docker://busybox
 * docker://library/busybox
 * docker://registry-1.docker.io/library/busybox

schu:
Updated with unit tests.

> Also, does this introduce a delta in the generated image manifest of the final ACI, compared to previous behavior?

I compared the manifest for the following Docker URLs for master + PR:

* docker://busybox
* docker://library/busybox
* docker://registry-1.docker.io/library/busybox
* docker://docker.io/library/busybox (only with PR)

The only difference is in `appc.io/docker/originalname`.

iaguis:
AFAICS, `appc.io/docker/originalname` is the same as before except that for `docker://docker.io/library/busybox` it is `docker.io/library/busybox` and before this wasn't working at all.

I'm fine with merging this. Any other concerns, @lucab?

lucab:
@iaguis Not really. I was asking about the originalname because rkt relies on that to avoid keeping multiple converted ACIs for the same docker image when shorthand names are used.

schu:
That's what I see for `appc.io/docker/originalname` for **both** master and PR version:

```
docker2aci docker://busybox
=> appc.io/docker/originalname == busybox
docker2aci docker://library/busybox
=> # appc.io/docker/originalname == library/busybox
docker2aci docker://docker.io/library/busybox                  # doesn't work with master
=> appc.io/docker/originalname == docker.io/library/busybox
docker2aci docker://registry-1.docker.io/library/busybox
=> appc.io/docker/originalname == registry-1.docker.io/library/busybox
```

Should it be the same `appc.io/docker/originalname` for all `docker://` URLs?

lucab:
@schu no, that delta is fine.

schu:
Ok, thanks!

Issue #259, opened 2018-02-26
Error: conversion error: error generating ACI: archive/tar: cannot encode header


zhsj:
While running `docker2aci -debug docker://docker.elastic.co/kibana/kibana:6.1.3` it shows:

```
Error: conversion error: error generating ACI: archive/tar: cannot encode header: Format specifies USTAR; and USTAR cannot encode Linkname="rootfs/var/lib/yum/yumdb/a/6f8ba1fc41ecc5f60aac5b4b7aa28d2566c55481-audit-libs-2.7.6-3.el7-x86_64/from_repo"
```

The error comes from

https://github.com/appc/docker2aci/blob/365911f6aa7f0a9a324a27bb6da0f7755df43041/lib/internal/internal.go#L632

`t.Header.Format` is `USTAR` here. The call to `tar.WriteHeader` triggers the error.

`t.Header` is parsed from 

https://github.com/appc/docker2aci/blob/365911f6aa7f0a9a324a27bb6da0f7755df43041/lib/internal/tarball/walk.go#L30

which is also from Go std lib `archive/tar`.

So, I think the fault is not `docker2aci`. But should `dokcer2aci` do some tricks to work around this?

One approach is to unset the `t.Header.Format`, at least it works for me currently.

```
diff --git a/lib/internal/internal.go b/lib/internal/internal.go
index 1f811fc..d694c39 100644
--- a/lib/internal/internal.go
+++ b/lib/internal/internal.go
@@ -609,6 +609,7 @@ func writeACI(layer io.ReadSeeker, manifest schema.ImageManifest, curPwl []strin
        fileMap := make(map[string]struct{})
        var whiteouts []string
        convWalker := func(t *tarball.TarFile) error {
+               t.Header.Format = tar.FormatUnknown
                name := t.Name()
                if name == "./" {
                        return nil
```

Any suggestions?

zhsj:
The previous comment is false, the error is still in appc/docker2aci

https://github.com/appc/docker2aci/blob/365911f6aa7f0a9a324a27bb6da0f7755df43041/lib/internal/internal.go#L629

Here, the `len(Linkname) > 100`, making the `t.Header.Format` which is USTAR invalid.

lucab:
How did you build this `docker2aci` binary? I'm starting to suspect that this is a regression in Go 1.10. If I try with the binary attached to the github release or with one locally build with go-1.9, I can't seem to reproduce this.

zhsj:
[...]

PR #260, opened 2018-02-26
lib/internal: fix tar header format

zhsj:
fix tar header format after manipulating the Name and Linkname.

Fixes: https://github.com/appc/docker2aci/issues/259

### DIFF
--- a/lib/internal/internal.go
+++ b/lib/internal/internal.go
@@ -613,23 +613,40 @@ func writeACI(layer io.ReadSeeker, manifest schema.ImageManifest, curPwl []strin
 		if name == "./" {
 			return nil
 		}
-		t.Header.Name = path.Join("rootfs", name)
-		absolutePath := strings.TrimPrefix(t.Header.Name, "rootfs")
+		newName := path.Join("rootfs", name)
+		absolutePath := strings.TrimPrefix(newName, "rootfs")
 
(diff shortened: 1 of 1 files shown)


zhsj:
The Format field is exposed since Go 1.10, should the CI bump its Go version?

zhsj:
As long as it uses `archive/tar`, I don't find other way to workaround. But if you don't want to enforce docker2aci depending go1.10, I can change the PR to use build profile(like separating the fixture to a function, and only building that function on go1.10), but it looks ugly...

lucab:
@iaguis any feedback? I would typically wait till 1.9 goes out of support (i.e. August 2018), but at the same time this seems worth fixing in a shorter term.

zhsj:
Reading the reference issue on https://github.com/coreos/bugs/issues/2402, it seems docker2aci on CoreOS 1688.5.3 not affected by this. However I don't use CoreOS, not sure what version the docker2aci is uses. If this is a regression in docker2aci, then we can fix it without bumping Go dependency to 1.10.

zhsj:
Reading the reference issue on https://github.com/coreos/bugs/issues/2402, it seems docker2aci on CoreOS 1688.5.3 not affected by this. However I don't use CoreOS, not sure what version the docker2aci is uses. If this is a regression in docker2aci, then we can fix it without bumping Go dependency to 1.10.

zhsj:
As suggested in the issue of Go, I changed the PR to copy Header struct.

PR #261, opened 2018-04-13
tests: update script to run within gopath

lucab:
This updates the test script to setup and move into gopath before
running any go command, as recent toolchains (>= 1.10) requires
to be within a valid gopath for proper dependency resolution.


iaguis (approved):
lgtm

